PR #1, opened 2017-03-01
Initial dump of my code

mcharsley:
Initial dump of my code. Supports
 brackets
 commas
-o
-mtime/ctime/atime
-print
-name/iname
-type (just file and directory at the moment)
-newer
-prune
-maxdepth/mindepth
-depth
-true/-false
And has a reasonable amount of unit/integration tests.

Quick summary of the files:
 - find/main.rs is an absolutely minimal wrapper that creates some dependencies and then calls into the library
 - find/mod.rs is the top-level library that handles argument parsing and walking down the filesystem tree
 - find/matchers/mod.rs defines a Matcher trait that takes a DirEntry and says whether it matches an arbitrary criteria or not, and has various associated helper functions 
 - find/matchers/logical.rs has some Matcher implementations for ANDing and ORing collections of sub-matchers
 - find/matchers/*.rs the other files in this submodule are fairly obvious, containing matchers that check name, size, age etc.
 - test_data/* has some static files that are used by the tests to reduce the need for creating and tearing down temporary files for all the tests -  I really wish Rust had a way to target a fake in-memory filesystem, like Java 7's FileSystem interface :-(

I'll leave that for you to review, and in the meantime, I'd better get on with perf and polishing my Google Resume...


Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[dependencies]
+glob = "0.2"
+# Current release of walkdir doesn't support depth-first operation. While we 
+# wait for https://github.com/BurntSushi/walkdir/pull/19 to be accepted and a
+# new release made, let's access my branch directly.
+walkdir = { git = "https://github.com/mcharsley/walkdir", rev = "dffefcf8db97a331a0f81d120e8aa20c1b36251e" }
+tempdir = "0.3"
+regex = "0.2"
+num-traits = "0.1.37"
```
antiagainst: Why requiring a patch number for `num-traits` but not others?

src/find/main.rs
```diff
@@ -0,0 +1,9 @@
+extern crate glob;
```
antiagainst: These source files should have copyright headers.

src/find/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+
+use std::cell::RefCell;
+use std::error::Error;
+use std::io::{Write, stderr, stdout};
+use std::rc::Rc;
+use std::time::SystemTime;
+use walkdir::WalkDir;
+use walkdir::WalkDirIterator;
+
+pub struct Config {
```
antiagainst: `#derive[Default]` maybe? It's meaningful to have a default config.

src/find/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+}
+
+/// The result of parsing the command-line arguments into useful forms.
+struct ParsedInfo {
+    matcher: Box<self::matchers::Matcher>,
+    paths: Vec<String>,
+    config: Config,
+}
+
+/// Function to generate a ParsedInfoi from the strings supplied on the command-line.
```
antiagainst: s/`ParsedInfoi`/`ParsedInfo`/

src/find/matchers/logical_matchers.rs
```diff
@@ -0,0 +1,477 @@
...
+pub struct AndMatcherBuilder {
+    submatchers: Vec<Box<Matcher>>,
+}
+
+impl AndMatcherBuilder {
+    pub fn new() -> AndMatcherBuilder {
+        AndMatcherBuilder { submatchers: Vec::new() }
+    }
+
+    pub fn new_and_condition(&mut self, matcher: Box<Matcher>) {
```
antiagainst: I would suggest just to call this as `and`. It's already concise and clear.

src/find/matchers/logical_matchers.rs
```diff
@@ -0,0 +1,477 @@
...
+
+impl OrMatcherBuilder {
+    pub fn new_and_condition(&mut self, matcher: Box<Matcher>) {
+        // safe to unwrap. submatchers always has at least one member
+        self.submatchers.last_mut().unwrap().new_and_condition(matcher);
+    }
+
+    pub fn new_or_condition(&mut self, arg: &str) -> Result<(), Box<Error>> {
+        if self.submatchers.last().unwrap().submatchers.is_empty() {
+            return Err(From::from(format!("invalid expression; you have used a binary operator \
```
antiagainst: I would suggest to define our own `Error` enum so that this can be an enumerant. It's more structured that way. :)

src/find/matchers/logical_matchers.rs
```diff
@@ -0,0 +1,477 @@
...
+        self.submatchers.last_mut().unwrap().new_or_condition(arg)
+    }
+
+    pub fn check_new_and_condition(&mut self) -> Result<(), Box<Error>> {
+        {
+            let child_or_matcher = &self.submatchers.last().unwrap();
+            let grandchild_and_matcher = &child_or_matcher.submatchers.last().unwrap();
+
+            if grandchild_and_matcher.submatchers.is_empty() {
+                return Err(From::from("invalid expression; you have used a binary operator '-a' \
```
antiagainst: And this can be another error enumerant. Something for the rest.

src/find/matchers/mod.rs
```diff
@@ -0,0 +1,732 @@
+mod logical_matchers;
```
antiagainst: Maybe considering make some of these modules as public (also in other modules) so we can actually call the functionality as a library? But definitely can wait.

src/find/matchers/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+    }
+}
+
+
+/// Builds a single AndMatcher containing the Matcher objects corresponding
+/// to the passed in predicate arguments.
+pub fn build_top_level_matcher(args: &[&str],
+                               config: &mut Config)
+                               -> Result<Box<Matcher>, Box<Error>> {
+    let (_, top_level_matcher) = try!(build_matcher_tree(args, config, 0, false));
```
antiagainst: Can use `?` for `try!`. But not a big thing. :)

src/find/matchers/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+                if i >= args.len() - 1 {
+                    return Err(From::from(format!("missing argument to {}", args[i])));
+                }
+                let file_time_type = match args[i] {
+                    "-atime" => time::FileTimeType::Accessed,
+                    "-ctime" => time::FileTimeType::Created,
+                    "-mtime" => time::FileTimeType::Modified,
+                    // This shouldn't be possible. We've already checked the value
+                    // is one of those three values.
+                    _ => panic!("Encountered unexpected value {}", args[i]),
```
antiagainst: `unreachable!()`?

src/find/matchers/size.rs
```diff
@@ -0,0 +1,174 @@
...
+    Byte,
+    TwoByteWord,
+    Block,
+    KiloByte,
+    MegaByte,
+    GigaByte,
+}
+
+impl Unit {
+    fn from_string(s: &str) -> Result<Unit, Box<Error>> {
```
antiagainst: Maybe `impl FromStr for Unit`?

src/find/matchers/type_matcher.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use find::matchers::{Matcher, MatcherIO};
+
+/// This matcher checks the type of the file.
+pub struct TypeMatcher {
+    file_type_fn: fn(&FileType) -> bool,
+}
+
+impl TypeMatcher {
+    pub fn new(type_string: &str) -> Result<TypeMatcher, Box<Error>> {
```
antiagainst: `impl FromStr for TypeMatcher`?

Cargo.toml
raphlinus: Also, given that we are building binaries, shouldn't we check in a `Cargo.lock` as well?

src/find/main.rs
raphlinus: Also, the copyright in LICENSE should be to either "Google Inc." or to "the findutils project maintainers".

src/find/matchers/size.rs
```diff
@@ -0,0 +1,174 @@
...
+use walkdir::DirEntry;
+
+use find::matchers::{ComparableValue, Matcher, MatcherIO};
+
+#[derive(Clone, Copy, Debug)]
+enum Unit {
+    Byte,
+    TwoByteWord,
+    Block,
+    KiloByte,
```
raphlinus: I could probably answer this myself from looking at the rest of the code, but should it be "Kilo" or "Kibi"?

src/find/matchers/size.rs
```diff
@@ -0,0 +1,174 @@
...
+        Unit::Block => 512,
+        Unit::KiloByte => 1024,
+        Unit::MegaByte => 1024 * 1024,
+        Unit::GigaByte => 1024 * 1024 * 1024,
+    };
+    // Skip pointless arithmetic.
+    if divisor == 1 {
+        return byte_size;
+    }
+    // We want to round up (e.g. 1 byte - 1024 bytes = 1k.
```
raphlinus: Another way to express this (which would avoid the need to special-case 0) is `(byte_size + divisor - 1) / divisor`. For better efficiency, might be best to use lg2(divisor), and shift right by that.

src/find/matchers/size.rs
```diff
@@ -0,0 +1,174 @@
...
+        let equal_to_1_blocks = SizeMatcher::new(ComparableValue::EqualTo(1), "b").unwrap();
+        let deps = FakeDependencies::new();
+
+        assert!(!equal_to_2_blocks.matches(&file_info, &mut deps.new_matcher_io()),
+                "512-byte file should not match size of 2 blocks");
+        assert!(equal_to_1_blocks.matches(&file_info, &mut deps.new_matcher_io()),
+                "512-byte file should match size of 1 block");
+    }
+
+
```
raphlinus: too many blank lines?

src/find/matchers/time.rs
```diff
@@ -0,0 +1,325 @@
+use std;
+use std::error::Error;
+use std::fs::{File, Metadata};
+use std::io::{stderr, Write};
+use std::time::SystemTime;
+use walkdir::DirEntry;
+
+use find::matchers::{ComparableValue, Matcher, MatcherIO};
+
+const SECS_PER_DAY: i64 = 60 * 60 * 24;
```
raphlinus: I'd write out "SECONDS"

src/find/matchers/printer.rs
```diff
@@ -0,0 +1,50 @@
+use walkdir::DirEntry;
+
+use find::matchers::{Matcher, MatcherIO};
+
+/// This matcher just prints the name of the file to stdout.
+pub struct Printer {
```
raphlinus: This can be written `pub struct Printer` (no braces).

src/find/matchers/prune.rs
```diff
@@ -0,0 +1,49 @@
+use walkdir::DirEntry;
+
+use find::matchers::{Matcher, MatcherIO};
+
+/// This matcher checks the type of the file.
+pub struct PruneMatcher {
```
raphlinus: Same comment as PrintMatcher.

src/find/matchers/time.rs
```diff
@@ -0,0 +1,325 @@
...
+        // duration_since will return an error.
+        Ok(self.given_modification_time.duration_since(this_time).is_err())
+    }
+}
+
+impl Matcher for NewerMatcher {
+    fn matches(&self, file_info: &DirEntry, _: &mut MatcherIO) -> bool {
+        match self.matches_impl(file_info) {
+            Err(e) => {
+                writeln!(&mut stderr(),
```
raphlinus: Consider a `print_err!` macro for error printing; among other things, this would make it easier to abstract the actual error logging.

src/find/matchers/time.rs
```diff
@@ -0,0 +1,325 @@
...
+        // wrapped in an error if now < this_time.
+        let age = match now.duration_since(this_time) {
+            Ok(duration) => duration,
+            Err(e) => {
+                is_negative = true;
+                e.duration()
+            }
+        };
+        let age_in_seconds: i64 = age.as_secs() as i64 * if is_negative { -1 } else { 1 };
+        // rust division truncates towards zero (see
```
raphlinus: This logic isn't exactly truncation, it's "age in days" - 1 if the age is an exact multiple. Maybe that's ok, but comment doesn't quite match code.

src/find/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+        paths: paths,
+        config: config,
+    })
+}
+
+fn process_dir<'a>(dir: &str,
+                   config: &Config,
+                   deps: &'a Dependencies<'a>,
+                   matcher: &Box<matchers::Matcher>)
+                   -> Result<i32, Box<Error>> {
```
raphlinus: Any particular reason this is i32, as opposed to, say, isize?

src/find/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+    }
+    Ok(found_count)
+}
+
+fn print_help() {
+    println!("Usage: find [path...] [expression]
+
+If no path is supplied then the current working directory is used by default.
+
+Early alpha implementation. Currently the only expressions supported are
```
raphlinus: Is this comment stale? It seems like you have more matchers defined.

src/find/matchers/logical_matchers.rs
```diff
@@ -0,0 +1,477 @@
...
+use walkdir::DirEntry;
+
+use find::matchers::{Matcher, MatcherIO};
+
+/// This matcher contains a collection of other matchers. A file only matches
+/// if it matches ALL the contained sub-matchers. For sub-matchers that have
+/// side effects, the side effects occur in the same order as the sub-matchers
+/// were pushed into the collection.
+pub struct AndMatcher {
+    submatchers: Vec<Box<Matcher>>,
```
raphlinus: Vec<Matcher> should work ok here, not sure why you're adding an additional `Box` indirection.

src/find/matchers/logical_matchers.rs
```diff
@@ -0,0 +1,477 @@
...
+        for x in self.submatchers {
+            submatchers.push(x.build());
+        }
+        Box::new(ListMatcher::new(submatchers))
+    }
+}
+
+
+/// A simple matcher that always matches.
+pub struct TrueMatcher {
```
raphlinus: Can also be unit-like struct.

src/find/matchers/logical_matchers.rs
```diff
@@ -0,0 +1,477 @@
...
+        true
+    }
+
+    fn has_side_effects(&self) -> bool {
+        false
+    }
+}
+
+/// A simple matcher that never matches.
+pub struct FalseMatcher {
```
raphlinus: Can also be unit-like struct.

mcharsley:
Thanks all. As a beginner with git, what's the procedure here? Do I make
changes to my tree and create a new pull request? Or is there something
clever I can do with the existing one?

Thanks

Mark

raphlinus:
The usual flow on github is to do an additional commit with the new changes and push that to your branch (mcharsley:master). That'll automatically update the pull request. Then one of us clicks "merge pull request" in the web UI when we approve.

antiagainst:
I would typically prefer "squash & merge" since that hides rounds of iteration details (like fixing typo, addressing comments, significant or not) and make the history more clean. :)

raphlinus:
I actually prefer squash and merge (it's much closer to gerrit-style workflows), but almost none of the github-hosted projects I work on use it.

antiagainst:
@raphlinus: Haha, we did that for all the projects we maintain: https://github.com/google/shaderc and https://github.com/KhronosGroup/SPIRV-Tools. ;-P

Cargo.toml
antiagainst: +1. That's recommended by cargo.

src/find/matchers/size.rs
antiagainst: I just wonder whether rustfmt will do the work to remove this.

.settings/com.github.rustdt.ide.core.prefs
```diff
@@ -0,0 +1,2 @@
+build_targets=<?xml version\="1.0" encoding\="UTF-8" standalone\="no"?>\n<build_targets xmlns\="com.github.rustdt.ide.core">\n<target auto_enabled\="false" config\="build" n_enabled\="true" version2\="true"/>\n<target auto_enabled\="true" config\="check" n_enabled\="false" version2\="true"/>\n</build_targets>\n
```
mcharsley: That shouldn't have been commited. I'll remove it and git-ignore it.

Cargo.toml
mcharsley: That was what num-traits documentation recommended using. As it turns out I don't need num-traits any more, so I'm deleting it.

mcharsley: Will add Cargo.lock and stop ignoring it

src/find/main.rs
mcharsley: Copyright headers? Eurgh :-( Is there a standard template I should be copying from?

Rather more embarrassing copyright line in LICENSE now corrected.

src/find/mod.rs
mcharsley: I could, but I'm struggling to see the benefit. max_depth absolutely should not default to 0. I could change 
```
impl Config {
    fn new() -> Config {
```
to 
```
impl Default for Config {
    fn default() -> Config {
```

But from what I can see I'd then have to change
`let config = Config::new()`
to
`let config : Config = Default::default()`
Which is more to type and requires `Default` to be `use`d. Happy to be persuaded if you can show me how it would be useful (I'm still learning the language after all)

src/find/mod.rs
mcharsley: Done

src/find/matchers/logical_matchers.rs
mcharsley: This was originally called "and" and OrMatcherBuilder's equivalent was originally called "or". But then ListMatcherBuilder didn't have an obvious matching name... I'd like to keep the 3 methods fairly consistent name-wise but am open to suggestions

src/find/matchers/logical_matchers.rs
mcharsley: Can I add that as a TODO for a later CL?

src/find/matchers/mod.rs
mcharsley: I prefer try! - my eye skips over ? and I miss the implied possible early return. But my eye is inexperienced at rust, so if ? is the standard idiom then I can switch... Let me know

src/find/matchers/mod.rs
mcharsley: Oooh: nice!

src/find/matchers/size.rs
mcharsley: Once again: nice!

src/find/matchers/size.rs
mcharsley: Done

src/find/matchers/size.rs
mcharsley: Normally I'd expect the compiler to optimise the division into right shifts, but this is getting sufficiently removed from just dividing by a constant that I'd understand if it failed.

Adding the divisor could potentially cause an overflow error for a spectacularly large file and short-cutting the entire function for 0 seems reasonable enough. If you disagree about the cost of pipeline stalls caused by conditionals and the relative risk of encountering exabyte-scale files I can change it though,

src/find/matchers/size.rs
mcharsley: My old nemesis: vertical whitespace :-(

Alas rustfmt doesn't help here (My eclipse workspace runs it at every save)

src/find/matchers/time.rs
mcharsley: Done

src/find/matchers/printer.rs
mcharsley: done

src/find/matchers/prune.rs
mcharsley: done

src/find/matchers/time.rs
mcharsley: Can I leave that as a TODO (as you say abstracting out error reporting will be useful)?

src/find/matchers/time.rs
mcharsley: Have changed the comment. Hope it's more clear now.

src/find/mod.rs
mcharsley: Force of habit. Will change to a u64 just to be safe.

src/find/mod.rs
mcharsley: It is indeed hopelessly out of date. Fixing now

src/find/matchers/logical_matchers.rs
mcharsley: done

src/find/matchers/logical_matchers.rs
mcharsley: done

mcharsley:
OK changes committed and pushed. Hopefully you can see them

Mark

mcharsley:
I'm a big fan of dependency injection. Less of a fan of people being
"clever" with Spring and Guice though.

Mark

src/find/mod.rs
antiagainst: Sorry, my bad. Not `#derive[Default]` but `impl Default`. You can have both and in `impl Default` just call `new`. So used to write `#derive[...]`. :(

The benefit is that you can enable clients of `Config`, e.g., a struct having a `Config` as field, to be able to use `#derive[Default]`. https://doc.rust-lang.org/core/default/trait.Default.html#derivable

Another usage could be possible, say if I'm writing my configuration in a json file and then using serde to load that json file into some data structure like:

```rust
struct MyGigantConfig {
    #[serde(default)]
    findutils_config: findutils::..::Config
    // Some other configs
}
```
Which means just use the default value for `Config` if the json file doesn't specify it.

src/find/matchers/logical_matchers.rs
antiagainst: That's fair.

src/find/matchers/logical_matchers.rs
antiagainst: Of course!

src/find/matchers/mod.rs
antiagainst: Both are fine, I think.

src/find/mod.rs
mcharsley: Fair enough. And I realised I can just use `let config = Config::default()` which is less painful than `let config : Config = Default::default()`

src/find/matchers/type_matcher.rs
mcharsley: I prefer the consistency of every Matcher having a new method (and a new_box), rather than ones that take a single string using FromStr and all the others using new.

There's a difference between "this is an object that can be represented as a string" and "this is an object that happens to take a single string as an argument".

antiagainst: It's not selecting one between `impl FromStr` and `pub fn new()`. Can have both and actually call `from_str()` in `new()`.

From the [spec](https://doc.rust-lang.org/std/str/trait.FromStr.html) of `FromStr`:

> A trait to abstract the idea of creating a new instance of a type from a string.

IMHO, that's what we are doing here, it's not "an object that can be represented as a string", which sounds more like `impl Display` for me.

Although I don't have strong examples to show the importance of having `impl FromStr`. It's basically conventional stuff, IMHO. So, entirely up to you. :)

antiagainst: It's not selecting one between `impl FromStr` and `pub fn new()`. Can have both and actually call `from_str()` in `new()`.

From the [spec](https://doc.rust-lang.org/std/str/trait.FromStr.html) of `FromStr`:

> A trait to abstract the idea of creating a new instance of a type from a string.

IMHO, that's what we are doing here, it's not "an object that can be represented as a string", which sounds more like `impl Display` for me.

Although I don't have strong examples to show the importance of having `impl FromStr`. It's basically conventional stuff, IMHO. So, entirely up to you. :)

src/find/matchers/time.rs
raphlinus: Sure, np.

src/find/matchers/mod.rs
raphlinus: My personal feeling is that ? is now standard and should be preferred. This is a point of style, though.

src/find/matchers/size.rs
raphlinus: The overflow can be solved by using .saturating_add() rather than simply +. It's a good point, though, you want to be very conscious of overflow behavior.

Because the divisor is dynamic, I wouldn't expect the optimization.

I don't have a strong feeling about any of this, just offering alternatives.

raphlinus:
I have no concerns that would block merging this, but won't merge right now in case Mark wants to do minor touchups based on the conversation. Feel free to merge.

antiagainst:
Also +1 from me. :)

mcharsley:
I've changed the try! macros to ? and added copyright headers. I'm happy for your guys to merge at your convenience :-)

Next step is probably -exec but I've got to spend time preparing for my SETI -> SWE transition first :-(

mcharsley:
Friendly ping. Can someone press the button to do the merge? Or am OK to do it? In which case, should I use
 - Create a merge commit
 - Squash and merge
 - Rebase and merge
 - something else?

thanks

mark

antiagainst:
Hi @mcharsley, definitely feel free to merge it! I'd recommend "squash and merge".

PR #2, opened 2017-03-08
Add Travis CI configuration

antiagainst:


.travis.yml
```diff
@@ -0,0 +1,20 @@
+language: rust
+
+sudo: false
+os:
+  - linux
+  - osx
```
mcharsley: Have you checked that the tests pass on a mac when run manually (I haven't)?

antiagainst: One of the test, `find_main_depth_first` fails on my mac. Will see that from Travis soon. :)

antiagainst: Oh, it also [fails on Linux](https://travis-ci.org/google/findutils/builds/209000807)...

mcharsley:
Gah! It works fine on my box :-(

Aha! was missing an "-ordered" flag. So contents of a given directory isn't
deterministic.

Expect a PR very soon

Mark

mcharsley:
I somehow managed to pull in the wrong direction so the fix is now in,
without any approvals :-(

Can you update your local copy and check that the tests now pass on other
machines?

thanks

Mark

antiagainst:
Rebased and pushed. Should trigger a new build job now.

The test still fails on my mac.

mcharsley:
Latest travis build is green:
https://travis-ci.org/google/findutils/builds/209483094, which includes
some runs on osx. Is find_main_depth_first still failing on your mac? Can
you check that it contains the sorted flag:

let rc = find_main(&["find", "./test_data/simple", "-sorted", "-depth"],
&deps);

thanks

Mark

antiagainst:
My fault. Somehow I failed to rebase this change against the current `master`.

antiagainst:
Merged. ;-P

PR #5, opened 2017-03-11
Fix some problems reported by clippy

antiagainst:


mcharsley:
LGTM. How do you run clippy by the way?

antiagainst:
Clippy needs a **nightly** compiler. Just need `cargo install clippy` and then `cargo clippy`.

PR #17, opened 2017-03-15
Added support for -exec[dir]

mcharsley:


mcharsley:
Friendly ping... It doesn't feel right to submit without an LGTM...

mcharsley:
Ah. Just noticed that I have to manually assign a reviewer <FX: embarrassed grin>

Cargo.toml
```diff
@@ -14,4 +14,8 @@ regex = "0.2"
 
 [[bin]]
 name = "find"
-path = "src/find/main.rs"
\ No newline at end of file
+path = "src/find/main.rs"
+
+[[bin]]
+name = "testing-commandline"
+path = "src/testing/commandline/main.rs"
```
raphlinus: Very minor nit: file is missing newline at end.

I'm wondering whether the testing-commandline should be made conditional, as it doesn't ship in production. The easiest way to do that would be to use a feature.

src/find/matchers/exec.rs
```diff
@@ -0,0 +1,95 @@
...
+
+use find::matchers::{Matcher, MatcherIO};
+
+pub struct SingleExecMatcher {
+    executable: String,
+    args: Vec<Option<OsString>>,
+    exec_in_parent_dir: bool,
+}
+
+
```
raphlinus: Extra newlines?

src/find/matchers/exec.rs
```diff
@@ -0,0 +1,95 @@
...
+use std::io::{Write, stderr};
+use std::path::Path;
+use std::process::Command;
+use walkdir::DirEntry;
+
+use find::matchers::{Matcher, MatcherIO};
+
+pub struct SingleExecMatcher {
+    executable: String,
+    args: Vec<Option<OsString>>,
```
raphlinus: So this is very much a style point, but Option doesn't feel like the right type for this; {} is not the absence of an arg, but the presence of the filename placeholder. I'd personally do `enum Arg { Filename, LiteralArg(OsString) }` and `args: Vec<Arg`>.

src/find/matchers/exec.rs
```diff
@@ -0,0 +1,95 @@
...
+
+
+
+impl SingleExecMatcher {
+    pub fn new(executable: &str,
+               args: &[&str],
+               exec_in_parent_dir: bool)
+               -> Result<SingleExecMatcher, Box<Error>> {
+
+        let transformed_args = args.iter()
```
raphlinus: The manpage for find says that only a single {} is allowed. I think this would be the correct place to check that.

src/find/matchers/mod.rs
```diff
@@ -57,11 +59,23 @@ pub trait Matcher {
...
+    /// the filename will be printed to stdout. While this is a compile-time
+    /// fact for most matchers, it's run-time for matchers that contain a
+    /// collection of sub-Matchers.
+    fn has_side_effects(&self) -> bool {
+        // most matchers don't have side-effects, so supply a default implementation.
+        return false;
+    }
+
+    /// Notification that find has finished processing a given directory.
+    fn finished_dir(&self, _: &PathBuf) {}
```
raphlinus: Personally I prefer giving a descriptive name to the variable and adding `#[allow(unused)]`. Alternatively, _descriptive_variable_name.

src/find/mod.rs
```diff
@@ -171,6 +171,7 @@ Early alpha implementation. Currently the only expressions supported are
  -atime [+-]N
  -mtime [+-]N
  -newer path_to_file
+ -exec[dir] executable [args] [{{}}] [more args];
```
raphlinus: I think there should be a space before the semicolon, emphasizing that it's a separate arg.

src/find/matchers/exec.rs
mcharsley: Fair point

src/find/matchers/exec.rs
mcharsley: http://pubs.opengroup.org/onlinepubs/009695399/utilities/find.html says that multiple {}s is implementation defined and GNU find allows multiple {}s. And it's more work to fail with multiple {}s than it is to support it.

Cargo.toml
mcharsley: I'll do that in a follow-up CL

src/find/matchers/exec.rs
mcharsley: Done

src/find/mod.rs
mcharsley: Done

src/find/matchers/mod.rs
mcharsley: Done

PR #20, opened 2017-03-20
Added support for -perm

mcharsley:


mcharsley:
NB this fixes #20

src/find/matchers/perm.rs
```diff
@@ -0,0 +1,480 @@
+// Copyright 2017 Google Inc.
+//
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+/// ! find's permission matching uses a very unix-centric approach, that would
```
raphlinus: First, I'm unfamiliar with the '!' syntax, did you mean "//!" ? Second, I get a feeling the comment is disconnected from something, maybe a #[cfg] attribute.

src/find/matchers/perm.rs
```diff
@@ -0,0 +1,480 @@
...
+        }
+    }
+
+    pub fn parse(string_value: &str) -> Result<(u32, ComparisonType), Box<Error>> {
+        // safe to unwrap as the regex is a compile-time constant.
+        let re = Regex::new("^([/-]?)([0-7]+)$").unwrap();
+
+        // have we been given a simple octal based string (e.g. /222)?
+        if let Some(m) = re.captures(string_value) {
+            // all these unwraps are safe because we checked the string in the regex above
```
raphlinus: Technically, this is not safe (in the sense of panic being impossible) because of overflow.

src/find/matchers/perm.rs
mcharsley: I did indeed mean //! So the comment applies to the entire module

src/find/matchers/perm.rs
mcharsley: Fair point

Issue #14, opened 2017-03-15
CodeHealth: fix windows support


mcharsley:
This should be fairly trivial with the code at time of writing: we just need to remove the hard-coded expectations about slashes being the path separator (mainly in the tests).

Some of the currently unimplemented features (user, group and symlinks) might have to involve some `if cfg!(windows)` magic.

On the symlink front, I'd advise take inspiration from https://github.com/BurntSushi/walkdir/blob/master/src/tests.rs on how to test interaction with symlinks on both unix and windows machines

mcharsley:
This will be fixed by #21

mcharsley:
Now fixed

PR #21, opened 2017-03-24
Fixed windows compatability

mcharsley:
This required
 - changing a lot of test code to supply/expect backslashes rather than
slashes when running on windows
  - tweaking the way testing-commandline reported its args (debug format
escapes backslashes when printing strings)
  - changing the NewerMatcher to get metadata from fs::metadata() rather
than from File's metadata method (the latter caused mysterious test
failures, which I didn't bother tracking down becasue the former is more
efficient anyway)
  - hiding more of perm.rs behind #[cfg(unix)] to stop "unused X"
warnings


mcharsley:
This should fix #14

src/find/matchers/perm.rs
```diff
@@ -271,8 +280,8 @@ impl PermMatcher {
     }
 
     #[cfg(not(unix))]
-    pub fn new(pattern: &str) -> Result<PermMatcher, Box<Error>> {
-        Err(From("Permission matching is not available on this platform"))
+    pub fn new(_dummy_pattern: &str) -> Result<PermMatcher, Box<Error>> {
```
antiagainst: I think you can use `_` directly here?

src/find/mod.rs
```diff
@@ -219,6 +219,18 @@ mod tests {
 
     use super::*;
 
+    #[cfg(windows)]
```
antiagainst: I just wonder whether there are existing libraries for doing this instead of manually implementing it.

src/find/matchers/perm.rs
mcharsley: In a previous review, Raph preferred more explicit names rather than a single underscore...

src/find/mod.rs
mcharsley: The proper way to do this is to use path::join and path::to_string.

But that's more verbose, and for *tests* I'm happy to use a simple search'n'replace bodge.

src/find/matchers/perm.rs
raphlinus: `_` is fine in an _implementation_ that doesn't use the parameter. But when defining a trait where the parameter will be used (even if it isn't in a default implementation) I do prefer a more explicit name.

mcharsley:
Ah. makes sense.

thanks

Mark

Issue #13, opened 2017-03-15
FeatureComplete: add -delete extension


mcharsley:
Of all GNU find's extensions, this is one of the most useful.

dahc:
This can be closed since PR #82 was merged.

sylvestre:
Thanks :)

PR #22, opened 2017-04-10
Implement -delete (#13)

Arcterus:
While -delete may be supported now, there are (as of yet) no tests.  I am planning to push another commit with tests, but if you want to merge it now that is fine as well (I am going to write tests anyway unless someone else does).

### DIFF
--- a/src/find/matchers/delete.rs
+++ b/src/find/matchers/delete.rs
@@ -7,39 +7,47 @@
  * file that was distributed with this source code.
  */
 
+use std::env;
 use std::io::{self, stderr, Write};
 use std::fs::{self, FileType};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use walkdir::DirEntry;
 
 use find::matchers::{Matcher, MatcherIO};
 
-pub struct DeleteMatcher;
+pub struct DeleteMatcher {
+    current_dir: PathBuf
+}
 
 impl DeleteMatcher {
-    pub fn new() -> DeleteMatcher {
-        DeleteMatcher
+    pub fn new() -> io::Result<DeleteMatcher> {
+        Ok(DeleteMatcher {
+            current_dir: env::current_dir()?
+        })
     }
 
-    pub fn new_box() -> Box<Matcher> {
-        Box::new(DeleteMatcher::new())
+    pub fn new_box() -> io::Result<Box<Matcher>> {
+        Ok(Box::new(DeleteMatcher::new()?))
     }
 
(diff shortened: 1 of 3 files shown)


mcharsley:
Looks good so far. Would prefer tests to be added before committing the changes.

In the meantime, I'll liase with Google's open source folks about the CLA/Googlebot thing. I'd assume we can drop that now uutils has taken the project over, but I'm not a lawyer...

Arcterus:
Yeah, that's not a problem.  I'll probably add them tomorrow as it's pretty late here.

Alright, keep me posted about that. :)

mcharsley:
The CLA webhook has now been removed :-)

src/find/matchers/delete.rs
```diff
@@ -0,0 +1,62 @@
...
+        Box::new(DeleteMatcher::new())
+    }
+
+    fn delete(&self, file_path: &Path, file_type: FileType) -> io::Result<()> {
+        if file_type.is_file() || file_type.is_symlink() {
+            fs::remove_file(file_path)
+        } else if file_type.is_dir() {
+            fs::remove_dir(file_path)
+        } else {
+            unimplemented!() // TODO: not sure what find does for block devices, etc.
```
tavianator: GNU find just calls `unlink()` (or `unlinkat()`) in this case, which will handle every file type except directories.  Looks like that's what `remove_file()` [does too](https://doc.rust-lang.org/std/fs/fn.remove_file.html#platform-specific-behavior), so presumably

```rust
if file_type.is_dir() {
    fs::remove_dir(file_path);
} else {
    fs::remove_file(file_path);
}
```

is fine

src/find/matchers/delete.rs
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+}
+
+impl Matcher for DeleteMatcher {
+    fn matches(&self, file_info: &DirEntry, _: &mut MatcherIO) -> bool {
+        match self.delete(file_info.path(), file_info.file_type()) {
```
tavianator: GNU find skips deleting `.`, so that `find -delete` (or `find . -delete`) doesn't fail.

sylvestre:
@Arcterus do you still want to land this? thanks

dahc:
@sylvestre If @Arcterus doesn't, I do.

What's blocking it at this point? Just missing tests? I'd be happy to put some together if needed (separate PR, I guess)

sylvestre:
@dahc Yeah, rebase (all files moved) + rustfmt + probably some clippy fixes + tests
(new PR is fine)

dahc:
@sylvestre new PR: #82

Issue #30, opened 2017-06-14
-help found too aggressively


tavianator:
```
$ ./target/debug/find -exec echo '{}' -help \;
Usage: find [path...] [expression]
...
```

But the `-help` should be an argument to `-exec`.

PR #34, opened 2017-06-30
	Fix for #30 -help found too aggressively

mcharsley:


tavianator:
Not quite the same behaviour as GNU find, not that it matters too much:

```
$ ./target/debug/find -help -asdf
Error: Unrecognized flag: '-asdf'
$ find -help -asdf
Usage: find [-H] [-L] [-P] [-Olevel] [-D help|tree|search|stat|rates|opt|exec] [path...] [expression]
...
```

PR #82, opened 2021-01-01
Implement -delete (#13) with tests (supersedes PR #22)

dahc:
This takes @Arcterus's implementation of -delete and fixes a few issues in hopes of getting it added.

The most significant change other than the tests is a bit of rework on skipping the working dir. The existing code tried to skip the working dir in general, which isn't quite right. Only a literal "." is supposed to be skipped, which is actually simpler to do.

Safely testing that behavior was a little tricky from the unit test perspective, so I took the liberty of adding some general integration tests, which I think provide some value on their own and are an easier context from which to test the dot thing.

Also included are a bunch of trivial compiler warning fixes (was rather noisy).

### DIFF
--- a/src/find/matchers/delete.rs
+++ b/src/find/matchers/delete.rs
@@ -8,8 +8,8 @@
  */
 
 use std::env;
-use std::io::{self, stderr, Write};
 use std::fs::{self, FileType};
+use std::io::{self, stderr, Write};
 use std::path::{Path, PathBuf};
 
 use walkdir::DirEntry;
@@ -17,13 +17,13 @@ use walkdir::DirEntry;
 use find::matchers::{Matcher, MatcherIO};
 
 pub struct DeleteMatcher {
-    current_dir: PathBuf
+    current_dir: PathBuf,
 }
 
 impl DeleteMatcher {
     pub fn new() -> io::Result<DeleteMatcher> {
         Ok(DeleteMatcher {
-            current_dir: env::current_dir()?
+            current_dir: env::current_dir()?,
         })
     }
 
@@ -50,10 +50,13 @@ impl Matcher for DeleteMatcher {
         match self.delete(path, file_info.file_type()) {
             Ok(_) => true,
             Err(f) => {
-                writeln!(&mut stderr(),
-                         "Failed to delete {}: {}",
-                         file_info.path().to_string_lossy(),
-                         f).unwrap();
+                writeln!(
+                    &mut stderr(),
+                    "Failed to delete {}: {}",
+                    file_info.path().to_string_lossy(),
+                    f
+                )
+                .unwrap();
                 false
             }
         }
@@ -65,6 +68,4 @@ impl Matcher for DeleteMatcher {
 }
 
 #[cfg(test)]
-mod tests {
-
-}
+mod tests {}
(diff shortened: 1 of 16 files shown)


sylvestre:
Excellent, thanks
I forgot that we don't have CI here :/
I will see if I can deploy something for this

sylvestre:
(or you can give it a try if motivated :)

dahc:
What sort of deploy? I'm not sure I follow

sylvestre:
we have appveyor with only windows:
https://ci.appveyor.com/project/Arcterus/findutils/builds/37069007
We should have more stuff like for coreutils:
https://github.com/uutils/coreutils/runs/1630985894

https://docs.github.com/en/free-pro-team@latest/actions/quickstart

tests/find_cmd_tests.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 Chad Williamson
+// Copyright 2021 Chad Williamson <chad@dahc.us>
```
sylvestre, line 1: Ah ah, first copyright 2021 that I see :)

dahc:
Oh, I see, thanks. I could look into that if you want... seems like a separate thing from this PR though? I'm working from an Ubuntu box, so I can confirm the tests work there. There's a bunch of clippy stuff to fix in master -- I just tried to make sure I didn't add any new warnings.

sylvestre:
As you wish!  It would be great to have at least Linux. I don't want to be a pain (or force your hand to get this PR landed)

In a perfect world, we could have the same as coreutils :)
https://github.com/uutils/coreutils/blob/master/.github/workflows/CICD.yml

dahc:
No worries, I'll take a crack at it later today or tomorrow. If it gets unexpectedly gnarly, I may throw up my hands. :wink:

src/find/matchers/delete.rs
```diff
@@ -0,0 +1,102 @@
...
+impl Matcher for DeleteMatcher {
+    fn matches(&self, file_info: &DirEntry, _: &mut MatcherIO) -> bool {
+        let path = file_info.path();
+        let path_str = path.to_string_lossy();
+
+        // This is a quirk in find's traditional semantics probably due to
+        // POSIX rmdir() not accepting "." (EINVAL). std::fs::remove_dir()
+        // inherits the same behavior, so no reason to buck tradition.
+        if path_str == "." {
+            return false;
```
tavianator, line 47: `return true` to match other find implementations

dahc: Nice catch, thanks! It didn't occur to me to chain another matcher afterward to check that... too used to -delete being the end of the line, I guess :)

dahc:
Alrighty, I've added a basic CI workflow for tests on Linux, MacOS, and Windows, along with fmt and clippy checks, and fixed (almost) all the clippy warnings (ugh). The workflow doesn't seemed to have triggered on this pull request (`pull_request` is on, but might not take effect until after this initial merge... I'm not sure), but you can see a successful run on the branch here: https://github.com/dahc/findutils/actions/runs/456592309

If you want to go whole-hog and match coreutils, I suggest looking into the [workflow template feature for organizations](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/sharing-workflows-with-your-organization), which seems like it might work well for uutils (but needs someone who can create a .github repo in the org, so I can't help you there :wink:).

sylvestre:
This is excellent, many thanks!

PR #86, opened 2021-02-26
type_matcher.rs: add support for: l b c p s

ilius:
Now all file types except for D (door on Solaris) are supported.

This is not blocked by the support of -H -L -P flags.
Because the default behavior of GNU find is "Never follow symbolic links".

```
-P     Never follow symbolic links.  This is the default behaviour.  When find examines or prints information
              about  files, and the file is a symbolic link, the information used shall be taken from the properties
              of the symbolic link itself.
```

### DIFF
--- a/src/find/matchers/type_matcher.rs
+++ b/src/find/matchers/type_matcher.rs
@@ -8,6 +8,9 @@ use std::error::Error;
 use std::fs::FileType;
 use walkdir::DirEntry;
 
+#[cfg(unix)]
+use std::os::unix::fs::FileTypeExt;
+
 use super::{Matcher, MatcherIO};
 
 /// This matcher checks the type of the file.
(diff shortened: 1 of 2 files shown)


sylvestre:
Could you also please add tests? thanks

ilius:
I'm still trying to test sym links on Windows.

ilius:
Okay I added tests.

src/find/mod.rs
```diff
@@ -104,6 +104,7 @@ fn parse_args(args: &[&str]) -> Result<ParsedInfo, Box<dyn Error>> {
 }
 
 #[allow(clippy::borrowed_box)] // FIXME?
+#[allow(clippy::unnecessary_wraps)] // FIXME?
```
sylvestre, line 107: Why did you add this?

ilius: The ci failed with `clippy::unnecessary-wraps` error

src/find/matchers/type_matcher.rs
```diff
@@ -38,6 +45,15 @@ impl TypeMatcher {
         })
     }
 
+    pub fn is_symlink(file_type: &FileType) -> bool {
```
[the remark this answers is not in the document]
ilius: To check -H -L -P flags later. You want me to directly use `FileType.is_symlink`?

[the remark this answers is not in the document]
sylvestre: Yeah, No need for a function here

src/find/mod.rs
sylvestre: Please fix the issue into a separate PR
A PR should focus on one thing :)
thanks

ilius:
Updated.

sylvestre:
Sorry for the latency :)

PR #87, opened 2021-03-01
Fix clippy errors in src/find/mod.rs

ilius:
- Error `clippy::unnecessary-wraps` that is failing ci
- Error `clippy::borrowed-box` that was ignored


sylvestre:
thanks!

PR #120, opened 2021-12-17
Add support for GNU-compatible printf

refi64:
This adds initial support for GNU find-compatible printf syntax. As far
as I can tell, it should implement all the specifiers, with the
following exceptions:

- Using zero as padding instead of space: GNU find isn't consistent with
  which numeric format specifiers support this.
- Specifying precision (`%x.yz`): I believe sparseness (%S)
  and seconds-since-epoch (`%A@` / `%C@` / `%T@`) are the only ones that
  allow this, making it only partially useful.
- Printing SELinux contexts (`%Z`): requires some extra dependencies and
  isn't within our scope.

uucore is added as a dependency, because it has several utility
functions that we need, the complexity of which is high enough that
using the extra dependency is quite helpful.

This also stubs out many of the specifiers for Windows, simply because I
have no idea how these should work there. (The defaults chosen should be
"good enough" for the moment, though.)

<hr>

Okay, so some more meta notes on this:

- Right now, printing the filesystem doesn't work reliably without: https://github.com/uutils/coreutils/pull/2778
- I'm not sure if what I put for the copyright in the new file is correct, please point it out if not!
- Also not sure if it would be worth removing `printer.rs` in favor of just putting `Printf::new("%p").unwrap()` (which I believe should be identical) in the main file.


wlozano0collabora:
Very strange test failure, it doesn't seem to be related to this PR.

refi64:
oh oops I had fixed it locally but forgot to push, just did it now

sylvestre:
Would it be possible to have high level tests?
https://github.com/uutils/findutils/blob/master/tests/find_cmd_tests.rs
thanks

refi64:
@sylvestre I added some high-level tests in `src/find/mod.rs` with a lot of the others there, does that look okay?

sylvestre:
> @sylvestre I added some high-level tests in `src/find/mod.rs` with a lot of the others there, does that look okay?

could you please move them into the tests directory? I prefer to have all integration tests at the same place. thanks

refi64:
@sylvestre Ah I misunderstood, just moved it over and actually testing the `find` binary itself (not `find_main`).

refi64:
Okay got CI on Windows working again with the moved tests 😅 Should be good now. (I also pointed it to uucore from Git so that it has the mentioned fix for grabbing the filesystem (otherwise `%F` doesn't work).)

sylvestre:
thanks
Seems that the coverage could be improved:
https://codecov.io/github/uutils/findutils/commit/f9ab5847cf4d41bfc15ed9464218ec6915977641
do you think you could add a few more tests ?

refi64:
@sylvestre I added a few more tests, with that, though, most of the still-remaining red is things that can't easily be tested in CI due to a high dependence on host system config (e.g. filesystem type, block count)

refi64:
(and CI is passing now, again)

sylvestre:
fails on ubuntu:
```
---- find::matchers::printf::tests::test_printf_special_types stdout ----
thread 'find::matchers::printf::tests::test_printf_special_types' panicked at 'called `Result::unwrap()` on an `Err` value: Error { depth: 2, inner: Io { path: Some("/dev/shm/multipath"), err: Os { code: 13, kind: PermissionDenied, message: "Permission denied" } } }', src/find/matchers/mod.rs:433:47
```

refi64:
@sylvestre Well in retrospect, trying to use `/dev` here was a terrible idea 😅 new version uses some custom FIFO/socket files instead in order to test special file types

sylvestre:
thanks and sorry for the latency :(

PR #121, opened 2021-12-27
Add an initial implementation of xargs

### LINKED ISSUE #37: xargs please!
As mentioned in https://github.com/uutils/coreutils/issues/947 , a Rust implementation of xargs would be dope for faster, more secure systems.

refi64:
This includes much of the core xargs functionality, with the following
notable exceptions:

- Parallel execution (`-P`): This option currently just does
  nothing, that way anything that passes -P can at least run without a
  notable behavior shift (other than simply being slower).
- Replacement strings (`-I`): This can easily be worked around via an
  intermediate shell invocation (e.g. `xargs -L1 sh -c 'do-things-with
  $@' --`).
- EOF strings (`-E`): I've honestly never seen this actually used,
  though it would not be particularly difficult to implement given the
  current architecture.

Closes #37

**This has not been tested with Windows CI yet and also depends on #120, so I've marked it as a draft.**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -221,6 +221,7 @@ version = "0.2.0"
 dependencies = [
  "assert_cmd",
  "chrono",
+ "clap",
  "filetime",
  "glob",
  "once_cell",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ authors = ["uutils developers"]
 
 [dependencies]
 chrono = "0.4"
+clap = "2.34"
 glob = "0.3"
 walkdir = "2.3"
 tempfile = "3"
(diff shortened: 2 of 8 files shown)


refi64:
Looks like the failures from #120 were affecting here too, should be working here now as well (also fixed the clippy warnings)

sylvestre:
looks like it is working well
did you try to run the other testsuites?

refi64:
> did you try to run the other testsuites?

Well all of them work locally, so :crossed_fingers:

sylvestre:
i did this: 
https://github.com/uutils/findutils-tracking/ :)

sylvestre:
could you please rebase this one too?

refi64:
Just rebased, the GNU test suite is now running on this xargs so a few new tests will fail (since this doesn't support stuff like `-i`/`-I` like I mentioned previously) but hey, that means the test suite is working!

PR #126, opened 2022-01-13
Add support for regex matching

refi64:
This adds support for -regextype, -regex, and -iregex, using Oniguruma
to implement support for the Emacs, grep, POSIX basic, and POSIX
extended regex types.

<hr>

Meta note: I'm not sure if it would be preferred to have this behind a feature, since oniguruma *is* a C dependency, or if `regex` should be removed instead due to functionality overlap.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,16 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
+
+      # For bindgen: https://github.com/rust-lang/rust-bindgen/issues/1797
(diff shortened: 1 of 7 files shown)


refi64:
Looks like CI is finally passing, you know it's a good day when you hit a bug in an unrelated crate due to compiler flags :sweat_smile:

sylvestre:
could you please rebase this patch to compare with the other testsuite? thanks

refi64:
Done:

Screenshot from 2022-01-22 09-16-03

Issue #128, opened 2022-01-16
shows the differences with the other testsuites like in coreutils


sylvestre:
In coreutils, we have a great way to show progress / regression compared to the GNU testsuite

For example:
https://github.com/uutils/coreutils/runs/4833285375?check_suite_focus=true
![image](https://user-images.githubusercontent.com/733326/149680179-933ea77a-c21f-436e-bd53-f1ffef787640.png)

Implementation:
https://github.com/uutils/coreutils/blob/main/.github/workflows/GnuTests.yml

It would be great to have this here too.

sylvestre:
@refi64 would be interested in implementing this? :)
thanks

refi64:
Hmm sounds interesting, presumably this would be running against the GNU and BFS test suites?

sylvestre:
yeah :)
(would be easier to review your patches :)

PR #129, opened 2022-01-21
Add initial parts of automated compatibility tests

refi64:
This is the first half of the needed changes to set up automated
compatibility tests against GNU findutils and bfs, handling the uploads
of the build logs and JSON results. The workflow itself is heavily based
on the one from uutils/coreutils:

https://github.com/uutils/coreutils/blob/main/.github/workflows/GnuTests.yml

but with various cleanups & tweaks to better suit findutils.

This does *not* include the actual regression comparisons, because those
will only pass once archives of these files are up on the main branch.

Ref #128

<hr>

Like the commit message mentions, this is the first half of the changes, because [the second half](https://github.com/uutils/findutils/compare/main...refi64:compat-tests-2?expand=1) needs already-built artifacts available. i.e., this needs to be merged *first*, then once CI passes, the second PR can be created and merged in.

I *have* tested these changes in a fork, you can see the results here:

- [Successful run](https://github.com/refi64/findutils-ci-tests/actions/runs/1730647383)
- [Run with regressions](https://github.com/refi64/findutils-ci-tests/actions/runs/1730679638) (I removed `-print` in this one)

I presume with this out of the way, I'll begin to see some of the open PRs merged? :wink: (Worth noting that, combining the current PRs with some offline work I have for `-L`, the amount of passing tests for BFS's suite goes up from the current 98 to 109!)

### DIFF
--- a/.github/workflows/compat.yml
+++ b/.github/workflows/compat.yml
@@ -1,0 +1,141 @@
+on: [push, pull_request]
+
+name: GNU findutils compatibility tests
+
+jobs:
+  gnu-tests:
+    name: Run GNU findutils tests
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout findutils
+      uses: actions/checkout@v2
+      with:
+        path: findutils
+    - name: Checkout GNU findutils
+      uses: actions/checkout@v2
+      with:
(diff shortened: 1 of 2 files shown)


sylvestre:
This is terrific, thanks :)

PR #130, opened 2022-01-22
Add remaining portions of automated compatibility tests

### LINKED ISSUE #128: shows the differences with the other testsuites like in coreutils

refi64:
This adds the rest of the compatibility test workflow, in order to
automatically run against the GNU findutils & BFS test suites as part of
standard CI and compare the results to the latest from the 'main'
branch.

Closes #128

<hr>

Now the workflow has passed for the other MR, so :crossed_fingers:

### DIFF
--- a/.github/workflows/compat.yml
+++ b/.github/workflows/compat.yml
@@ -74,6 +74,42 @@ jobs:
       with:
         name: gnu-result
         path: gnu-result.json
+    - name: Download the result
+      uses: dawidd6/action-download-artifact@v2
+      with:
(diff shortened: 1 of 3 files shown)


sylvestre:
sweet, bravo :)

PR #131, opened 2022-01-22
Implement support for following symlinks

refi64:

<hr>

Unfortunately, it wasn't possible to make work as needed without these two PRs to walkdir: https://github.com/BurntSushi/walkdir/pull/159 https://github.com/BurntSushi/walkdir/pull/160

However, it seems that PRs to walkdir have been stagnant since late 2020 / early 2021, so for the purpose of testing, I created a single branch on my fork with both PRs merged in, and Cargo.toml in here is pointing to that branch. I'm not quite sure how to handle this as a result, so this is in draft status. Any thoughts on how to handle this?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -766,8 +766,7 @@ dependencies = [
 [[package]]
 name = "walkdir"
 version = "2.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "808cf2735cd4b6866113f648b791c6adc5714537bc222d9347bb203386ffda56"
+source = "git+https://github.com/refi64/walkdir?branch=symlinks#f799a1c6c008ecd9810f124237470435707ea253"
 dependencies = [
  "same-file",
  "winapi",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,9 @@ predicates = "2"
 serial_test = "0.5"
 tempfile = "3"
 
+[patch.crates-io]
+walkdir = { git = "https://github.com/refi64/walkdir", branch = "symlinks" }
+
 [[bin]]
 name = "find"
 path = "src/find/main.rs"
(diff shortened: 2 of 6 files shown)


sylvestre:
Clap clap:
```

Warning: Congrats! The bfs test test_L is now passing!
Warning: Congrats! The bfs test test_L_depth is now passing!
Warning: Congrats! The bfs test test_L_type_l is now passing!
Warning: Congrats! The bfs test test_flag_comma is now passing!
Warning: Congrats! The bfs test test_flag_weird_names is now passing!
Warning: Congrats! The bfs test test_type_l is now passing!
```

refi64:
Ah nice! This seems to affect quite a few tests. Any ideas for how to handle the walkdir issue, though?

mcharsley:
The initial release of findutils was pinned to a specific version in my
branch of walkdir, while I waited for
https://github.com/BurntSushi/walkdir/pull/19 to be accepted and included
in a release.

So it's not without precedent.

Long term, I'd suggest giving it a few months and then making our own
branch if the PRs keep piling up there untouched?

Mark

tavianator:
It is possible to work around https://github.com/BurntSushi/walkdir/pull/160 the same way `fd` does it: https://github.com/sharkdp/fd/blob/9f5ed8534e824b3e2a0934ff3aa0b781cb370688/src/walk.rs#L490-L503.  But it would be nice if walkdir would do it itself.

refi64:
@tavianator I was originally going to try using a similar system here, but there were 2 things that made me go with the walkdir patches instead:

- walkdir patches are needed anyway so that `find a-symlink` will not traverse into it without -L. The alternative would be to manually check if it's a symlink ourselves first but...
- ...findutils does seem to use more platform-specific `DirEntry` functionality that fd does, so re-implementing a custom `DirEntry` wrapper would be a decent bit messier, at least from my initial investigations

That being said, if the walkdir PR stays stagnant for too long, it may be worth switching over.

sylvestre:
no activity for a while, closing

PR #132, opened 2022-01-24
Update uucore to 0.0.12

refi64:
This version includes the needed filesystem type detection fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -244,15 +244,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "getopts"
-version = "0.2.21"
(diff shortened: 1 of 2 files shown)


refi64:
The test failure seems to be from something I never added to `xargs` still:

> `error: Found argument '-I' which wasn't expected, or isn't valid in this context`

so I'm not sure why it's only failing now

sylvestre:
maybe the upstream testsuite has been updated?

refi64:
Looks like this was due to #133; when xargs was merged in, the tests regressed, and now the workflows are stuck comparing to the pre-xargs version

PR #134, opened 2022-01-26
remove the unused tempfile dep

sylvestre:
cargo +nightly udeps


sylvestre:
udeps is wrong:

```
   Compiling findutils v0.2.0 (/home/runner/work/findutils/findutils)
error[E0463]: can't find crate for `tempfile`
Error:   --> src/lib.rs:12:1
   |
12 | extern crate tempfile;
   | ^^^^^^^^^^^^^^^^^^^^^^ can't find crate

```

tertsdiepraam:
Actually, I think udeps was sort of right. The crate is only used in tests and should probably be a dev-dependency instead of a normal dependency.

sylvestre:
right, fixed :)

sylvestre:
https://github.com/uutils/findutils/pull/135

PR #135, opened 2022-01-26
Move tempfile dep to dev-dep

sylvestre:
cargo +nightly udeps


sylvestre:
automerged as I would like to make a new release :)

PR #136, opened 2022-01-26
Add support for -xdev

refi64:
Unfortunately, the nature of this makes it incredibly hard to test for
in CI; not even BFS's massive test suite covers it. That being said,
it's a simple enough addition that the risk it incurs hopefully
shouldn't be too large.

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -392,6 +392,11 @@ fn build_matcher_tree(
                 config.depth_first = true;
                 None
             }
+            "-xdev" => {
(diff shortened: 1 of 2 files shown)


sylvestre:
For others:
```
       -xdev  Don't descend directories on other filesystems.
```

tavianator:
> not even BFS's massive test suite covers it

Actually `./tests.sh --sudo` does cover it, but it needs `sudo` access.  And it's probably not directly useful to uutils since `--sudo` also tests some bfs-specific extensions.  There's no way (yet) to ask for the `--sudo` tests that are GNU compatible.

PR #139, opened 2022-01-31
Avoid skipping the entire directory if a file hits -prune

refi64:
This matches up with the behavior of GNU findutils, as documented in its
man page:

> **if the file is a directory**, do not descend into it


sylvestre:
Looks good, thanks!

PR #141, opened 2022-02-01
matchers: Replace new_box() with an into_box() trait method

tavianator:


sylvestre:
could you please explain in the commit message the why ?
thanks

Issue #114, opened 2021-09-07
Address the differences with the GNU testsuite results


sylvestre:
Currently:
```
# TOTAL: 16
# PASS:  3
# SKIP:  1
# XFAIL: 0
# FAIL:  11
# XPASS: 0
# ERROR: 1
```

tavianator:
If you're looking for more external tests, I have a fairly large `find` testsuite as part of https://github.com/tavianator/bfs.  You can use it like this:

```
$ git clone https://github.com/tavianator/bfs.git
$ cd bfs
$ make all
...
$ ./tests.sh --bfs=/absolute/path/to/uutils/findutils/target/release/find --gnu
...
tests passed: 82
tests failed: 143
```

Note that currently GNU find scores

```
$ ./tests.sh --bfs=/usr/bin/find --gnu
...
tests passed: 206
tests failed: 19
```

due to things that I consider bugs in findutils.

sylvestre:
@tavianator tavianator oh, cool, thanks

You would be interested in adding this testsuite in:

https://github.com/uutils/findutils/blob/master/util/build-gnu.sh ?

tavianator:
@sylvestre Here you go: https://github.com/uutils/findutils/pull/116

tavianator:
I just noticed your run of the GNU tests is missing a lot of them.  A complete `make check` includes the dejagnu tests:

```
                === find Summary ===

# of expected passes            487
find (GNU findutils) 4.9.0.1-caba
...
                === xargs Summary ===

# of expected passes            96
xargs (GNU findutils) 4.9.0.1-caba
```

Currently uutils gets:

```
                === find Summary ===

# of expected passes            184
# of unexpected failures        624
find (Rust) 0.3.0
...
                === xargs Summary ===

# of expected passes            50
# of unexpected failures        46
xargs 0.3.0
```

PR #144, opened 2022-02-08
ci: Also run the dejagnu tests from GNU findutils

tavianator:
See https://github.com/uutils/findutils/issues/114#issuecomment-1027928458


sylvestre:
Seems that it failed:
https://github.com/uutils/findutils/runs/5103547383?check_suite_focus=true

tavianator:
> Seems that it failed: https://github.com/uutils/findutils/runs/5103547383?check_suite_focus=true

Yeah I can reproduce.  I'm thinking of moving a lot of the code out of compat.yml into `build-gnu.sh` so it's easier to test locally.

sylvestre:
Ah ah

`Warning: Changes from main: PASS +360 / FAIL +382 / ERROR +0 / SKIP +0 `

PR #163, opened 2022-05-25
ci: Update bfs testsuite to version 2.6

tavianator:
The latest bfs properly groups the --sudo tests, so we can test things
that require root like -xdev.


sylvestre:
`Warning: Changes from main: PASS +3 / SKIP +0 / FAIL +7`
I guess it is expected

PR #165, opened 2022-05-30
find/matchers: Implement -mount as an alias for -xdev

tavianator:


sylvestre:
would it be possible to add a test?

tavianator:
Hmm, there's no existing test for `-xdev` and writing a good one requires actually mounting something.  I could add a test that at least ensures that `-mount`/`-xdev` don't change the results on `test_data`.

sylvestre:
and improves the code coverage, cool :)

PR #166, opened 2022-05-30
find/matchers: Add the `ed` and `sed` regex types

tavianator:


sylvestre:
please also add a test to make sure we don't regress, thanks

PR #188, opened 2022-10-21
Fix a clippy warning

sylvestre:
`this expression borrows a value the compiler would automatically borrow`


src/find/matchers/printf.rs
```diff
@@ -206,7 +206,7 @@ impl FormatStringParser<'_> {
         if digits > 0 {
             // safe to unwrap: we already know all the digits are valid due to
             // the above checks.
-            Some((&start[0..digits]).parse().unwrap())
+            Some((start[0..digits]).parse().unwrap())
```
tavianator, line 209: The parentheses around `(start[0..digits])` are also redundant

PR #210, opened 2023-02-20
Add support for embedded "{}"

int3:
Given

    find . -exec foo{}bar \;

GNU find will replace `{}` by the filename. This commit matches that behavior.

Note that this is not required by the POSIX spec.

The GNU testsuite doesn't cover this case, so I've added a test in this repo.

I'm new to Rust, so please be gentle (:

### DIFF
--- a/src/find/matchers/exec.rs
+++ b/src/find/matchers/exec.rs
@@ -14,7 +14,7 @@ use walkdir::DirEntry;
 use super::{Matcher, MatcherIO};
 
 enum Arg {
-    Filename,
+    FileArg(Vec<OsString>),
     LiteralArg(OsString),
 }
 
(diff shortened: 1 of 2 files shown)


src/find/matchers/exec.rs
```diff
@@ -32,9 +32,14 @@ impl SingleExecMatcher {
...
-            .map(|&a| match a {
-                "{}" => Arg::Filename,
-                _ => Arg::LiteralArg(OsString::from(a)),
+            .map(|&a| {
+                let parts = a.split("{}").collect::<Vec<_>>();
+                if parts.len() == 1 {
+                    // No {} present
+                    Arg::LiteralArg(OsString::from(a))
+                } else {
+                    Arg::FileArg(parts.iter().map(|p| OsString::from(p)).collect())
```
sylvestre, line 41: clippy isn't happy here :)

int3: Whoops. Fixed :)

int3:
The failing jobs don't seem to be related to my changes

int3:
Did I bork the test runner somehow? Perhaps I shouldn't have force-pushed?

sylvestre:
I wonder if this isn't an issue with out CI.

sylvestre:
i am surprised that it didn't fix any GNU or BFS tests.
Could you please have a look?
thanks

tavianator:
So I can confirm that locally it fixes the `common/exec_substring` test.  Why doesn't that show up on CI?  I think it's because the action downloads the most recent `bfs-result.json` from a run on the `main` branch: https://github.com/uutils/findutils/actions/runs/4257476104/workflow#L119-L127

But if you look at the runs on the `main` branch: https://github.com/uutils/findutils/actions?query=branch%3Amain

you'll see it includes *this PR itself* since the branch is `int3:main`.  So somehow we need to tell `action-download-artifact` to only look at *our* main branch, not other branches that might be called `main`.

int3:
Would you like me to make a new PR with a different branch name?

int3:
Reopened with a different branch here: https://github.com/uutils/findutils/pull/213

PR #213, opened 2023-02-26
Add support for embedded "{}"

int3:
Given

  find . -exec foo{}bar \;

GNU find will replace `{}` by the filename. This commit matches that behavior.

Note that this is not required by the POSIX spec.

The GNU testsuite doesn't cover this case, so I've added a test in this repo.

### DIFF
--- a/src/find/matchers/exec.rs
+++ b/src/find/matchers/exec.rs
@@ -14,7 +14,7 @@ use walkdir::DirEntry;
 use super::{Matcher, MatcherIO};
 
 enum Arg {
-    Filename,
+    FileArg(Vec<OsString>),
     LiteralArg(OsString),
 }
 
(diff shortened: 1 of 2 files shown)


tavianator:
It still didn't make a difference for reporting new passing tests, since it's still comparing against the last `main` build which was the previous PR

int3:
Ah I see. Is that a blocker for merging this?

int3:
Ping?

sylvestre:
sorry, i would like to see more tests:
* find_cmd_tests.rs please add similar tests here 
* with some utf8 input
* with some failures (for example: if the exec doesn't exist)
thanks

int3:
> find_cmd_tests.rs please add similar tests here

You mean find_exec_tests.rs, right? Since this is an exec change

> with some utf8 input

Are there any tests that cover utf8 that I could reference?

> with some failures (for example: if the exec doesn't exist)

This seems kind of orthogonal to the PR -- regardless of whether there are multiple `{}` in a filename doesn't change whether the exec exists

int3:
Okay I found a bunch of utf8 tests in the coreutils repo but none in the findutils repo. I'm not against adding those tests but again they seem orthogonal to this PR

sylvestre:
ok, don't hesitate if you want to add them :)

Issue #226, opened 2023-04-02
warning: only a 'panic!' in 'if'-then statement


sylvestre:
10 occurrences found with
`cargo +nightly clippy  -- -W clippy::pedantic `
```
warning: only a `panic!` in `if`-then statement
  --> src/find/matchers/lname.rs:76:13
   |
76 | /             if e.kind() != ErrorKind::AlreadyExists {
77 | |                 panic!("Failed to create sym link: {:?}", e);
78 | |             }
   | |_____________^ help: try instead: `assert!(!(e.kind() != ErrorKind::AlreadyExists), "Failed to create sym link: {:?}", e);`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_assert
   = note: `-W clippy::manual-assert` implied by `-W clippy::pedantic`
```

PR #227, opened 2023-04-02
implemented assert! for resolving warning panic!

HelloShiv:
- Issue stated in [Issues #226](https://github.com/uutils/findutils/issues/226)
#226 

Used assert! statement instead of alert,  to indicate a failed condition without stopping the program immediately.

```
if e.kind() != ErrorKind::AlreadyExists {
    assert!(false, "Failed to create sym link: {:?}", e);
}
```

### DIFF
--- a/src/find/matchers/name.rs
+++ b/src/find/matchers/name.rs
@@ -48,13 +48,13 @@ mod tests {
         #[cfg(unix)]
         if let Err(e) = symlink("abbbc", "test_data/links/link-f") {
             if e.kind() != ErrorKind::AlreadyExists {
-                panic!("Failed to create sym link: {:?}", e);
+               assert!(false, "Failed to create sym link: {:?}", e);
             }
         }
         #[cfg(windows)]
         if let Err(e) = symlink_file("abbbc", "test_data/links/link-f") {
             if e.kind() != ErrorKind::AlreadyExists {
-                panic!("Failed to create sym link: {:?}", e);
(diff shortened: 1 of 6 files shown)


sylvestre:
you should follow the recommendation:
```
   | |_____________^ help: try instead: `assert!(!(e.kind() != ErrorKind::AlreadyExists), "Failed to create sym link: {:?}", e);
```

src/find/matchers/name.rs
```diff
@@ -47,15 +47,19 @@ mod tests {
     fn create_file_link() {
         #[cfg(unix)]
         if let Err(e) = symlink("abbbc", "test_data/links/link-f") {
-            if e.kind() != ErrorKind::AlreadyExists {
-                panic!("Failed to create sym link: {:?}", e);
-            }
+            assert!(
+                !(e.kind() != ErrorKind::AlreadyExists),
```
tertsdiepraam, line 51: There's a double negation here now that can be removed:
```suggestion
                e.kind() == ErrorKind::AlreadyExists,
```
Clippy probably would suggest the same thing.

tertsdiepraam (commented):
Nice! I think we can clean up just a bit more than what clippy suggests by removing the double negation.

src/find/matchers/name.rs
HelloShiv: implementing it

Issue #225, opened 2023-04-02
Fix clippy warning `unnecessary `!=` operation`


sylvestre:
```
warning: unnecessary `!=` operation
  --> src/testing/commandline/main.rs:81:26
   |
81 |           destination_dir: if args[1] != "-" {
   |  __________________________^
82 | |             Some(args[1].clone())
83 | |         } else {
84 | |             None
85 | |         },
   | |_________^
   |
   = help: change to `==` and swap the blocks of the `if`/`else`
```

PR #235, opened 2023-04-10
replaced not equal operation with equality operator

HelloShiv:
- [Issue stated in ](https://github.com/uutils/findutils/issues/225)    #225 

Implemented clippy warning suggestion

### DIFF
--- a/src/testing/commandline/main.rs
+++ b/src/testing/commandline/main.rs
@@ -78,7 +78,7 @@ fn main() {
         usage();
     }
     let mut config = Config {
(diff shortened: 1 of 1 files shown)


mcharsley (changes requested):
Doesn't the linked bug say to swap the if and else contents as well as changing != to ==?

HelloShiv:
> Doesn't the linked bug say to swap the if and else contents as well as changing != to ==?

My apologies, I forgot to include the swap part. Incooperating the swap part

mcharsley (approved):

Issue #310, opened 2024-02-11
Implement `-I / --replace`


sylvestre:
```
  -I R                         same as --replace=R
  -i, --replace[=R]            replace R in INITIAL-ARGS with names read
                                 from standard input, split at newlines;
                                 if R is unspecified, assume {}
```

DweepKM:
@sylvestre 
I would like to work on this issue.

sylvestre:
Could you please try another one ? I started this one

DweepKM:
No issue, let me know if there's any way I can assist you.

DweepKM:
@sylvestre 
Are you addressing issues #308  and #307  as well?

sylvestre:
No they are separate

PR #323, opened 2024-02-25
xargs: Implement replace / -I (Closes: #310)

sylvestre:
Closes: #310

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ nix = { version = "0.28", features = ["fs"] }
 predicates = "3"
 serial_test = "3.0"
 tempfile = "3"
+pretty_assertions = "1.4.0"
 
 [[bin]]
 name = "find"
--- a/src/xargs/mod.rs
+++ b/src/xargs/mod.rs
@@ -28,6 +28,7 @@ mod options {
     pub const MAX_PROCS: &str = "max-procs";
     pub const NO_RUN_IF_EMPTY: &str = "no-run-if-empty";
     pub const NULL: &str = "null";
+    pub const REPLACE: &str = "replace";
     pub const VERBOSE: &str = "verbose";
 }
 
@@ -40,6 +41,7 @@ struct Options {
     max_lines: Option<usize>,
     no_run_if_empty: bool,
     null: bool,
+    replace: Option<String>,
     verbose: bool,
 }
 
@@ -340,13 +342,14 @@ struct CommandBuilderOptions {
     limiters: LimiterCollection,
     verbose: bool,
     close_stdin: bool,
+    replace: Option<String>,
 }
-
 impl CommandBuilderOptions {
     fn new(
         action: ExecAction,
         env: HashMap<OsString, OsString>,
         mut limiters: LimiterCollection,
+        replace: Option<String>,
     ) -> Result<Self, ExhaustedCommandSpace> {
         let initial_args = match &action {
             ExecAction::Command(args) => args.iter().map(|arg| arg.as_ref()).collect(),
@@ -366,6 +369,7 @@ impl CommandBuilderOptions {
             limiters,
             verbose: false,
             close_stdin: false,
+            replace,
         })
     }
 }
@@ -398,14 +402,41 @@ impl CommandBuilder<'_> {
         };
 
         let mut command = Command::new(entry_point);
-        command
-            .args(initial_args)
-            .args(&self.extra_args)
-            .env_clear()
-            .envs(&self.options.env);
+
+        if let Some(replace_str) = &self.options.replace {
+            // we replace the first instance of the replacement string with
+            // the extra args, and then replace all instances of the replacement
+            let replacement = self
+                .extra_args
+                .iter()
+                .map(|s| s.to_string_lossy())
+                .collect::<Vec<_>>()
+                .join(" ");
+            let initial_args: Vec<OsString> = initial_args
+                .iter()
+                .map(|arg| {
+                    let arg_str = arg.to_string_lossy();
+                    OsString::from(arg_str.replace(replace_str, &replacement))
+                })
+                .collect();
+
+            command
+                .args(&initial_args)
+                .env_clear()
+                .envs(&self.options.env);
+        } else {
+            // don't do any replacement
+            command
+                .args(initial_args)
+                .args(&self.extra_args)
+                .env_clear()
+                .envs(&self.options.env);
+        };
+
         if self.options.close_stdin {
             command.stdin(Stdio::null());
         }
+
         if self.options.verbose {
             eprintln!("{command:?}");
         }
@@ -811,11 +842,21 @@ fn do_xargs(args: &[&str]) -> Result<CommandResult, XargsError> {
         )
         .arg(
             Arg::new(options::VERBOSE)
-                .short('t')
-                .long(options::VERBOSE)
-                .help("Be verbose")
-                .action(ArgAction::SetTrue),
+            .short('t')
+            .long(options::VERBOSE)
+            .help("Be verbose")
+            .action(ArgAction::SetTrue),
         )
+        .arg(
+            Arg::new(options::REPLACE)
+                .long(options::REPLACE)
+                .short('I')
+                .short_alias('i')
+                .num_args(0..=1)
+                .value_parser(clap::value_parser!(String))
+                .help("Replace R in INITIAL-ARGS with names read from standard input; if R is unspecified, assume {}"),
+        )
+
         .try_get_matches_from(args);
 
     let matches = match matches {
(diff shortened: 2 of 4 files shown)


tests/xargs_tests.rs
cakebaker, line 1: The tests look incorrect to me because GNU `xargs` doesn't have a `-R` option:
```bash
$ printf "foo" | xargs -R {} echo "{} bar"
xargs: invalid option -- 'R'
Try 'xargs --help' for more information.
```

sylvestre: Good point, i was confused. fixed!

cakebaker, line 1: I think there are some tests missing for when there is no value specified with `--replace` and `-i`. In both cases `{}` is used as default value:
```bash
$ printf "foo" | xargs --replace echo "{}bar"
foobar
$ printf "foo" | xargs -i echo "{}bar"
foobar
```

sylvestre: yeah, it isn't manageable with clap 2 for now  afaik :/

src/xargs/mod.rs
```diff
@@ -40,6 +41,7 @@ struct Options {
     max_lines: Option<usize>,
     no_run_if_empty: bool,
     null: bool,
+    replace: Option<String>,
```
tertsdiepraam, line 44: Maybe we should be using a `&[u8]` here? GNU probably accepts invalid utf-8. Otherwise, this is looking good.

tests/xargs_tests.rs
sylvestre: done with
                 .short_alias('i')

src/xargs/mod.rs
sylvestre: i just tried and it requires adding lifetime everywhere
is it worst it?

Cargo.toml
```diff
@@ -26,6 +26,7 @@ nix = { version = "0.28", features = ["fs"] }
 predicates = "3"
 serial_test = "3.0"
 tempfile = "3"
+pretty_assertions = "0.6.1"
```
cakebaker, line 29: Do you use this version on purpose? It's quite old, the current version is `1.4.0`.

sylvestre: bad copy and paste, fixed :)

tests/xargs_tests.rs
cakebaker, line 1: I think the changes to `xargs_exec_failure`, `xargs_exec_urgent_failure`, and `xargs_exec_with_signal` remove the most important part of those tests: ensuring the correct exit code is returned.

sylvestre: it is tested by this I think, no?

`    assert!(result.is_ok(), "xargs failed: {:?}", result);`

cakebaker: No, that line passes fine if I change the exit codes of `testing-commandline` :) And even if it would test it, in the test code it is no longer visible that the exit codes `123`, `124`, and `125` are expected for the different error situations.

src/xargs/mod.rs
```diff
@@ -813,11 +841,21 @@ fn do_xargs(args: &[&str]) -> Result<CommandResult, XargsError> {
...
-                .action(ArgAction::SetTrue),
+            .short('t')
+            .long(options::VERBOSE)
+            .help("Be verbose")
+            .action(ArgAction::SetTrue),
         )
+        .arg(
+            Arg::new(options::REPLACE)
+                .long(options::REPLACE)
+                .short('I')
```
cakebaker, line 852: I think `-I` has to be an `Arg` on its own because its behavior is slightly different from `-i`: `-I` requires a value  whereas `-i` doesn't:
```
$ printf "foo" | xargs -I echo {} _ bar
xargs: {}: No such file or directory
$ printf "foo" | xargs -I {} echo {} _ bar
foo _ bar

$ printf "foo" | xargs -i echo {} _ bar
foo _ bar
```

src/xargs/mod.rs
```diff
@@ -813,11 +841,21 @@ fn do_xargs(args: &[&str]) -> Result<CommandResult, XargsError> {
         )
         .arg(
             Arg::new(options::VERBOSE)
-                .short('t')
-                .long(options::VERBOSE)
-                .help("Be verbose")
-                .action(ArgAction::SetTrue),
```
cakebaker, line 819: This is the correct indentation.

src/xargs/mod.rs
```diff
@@ -813,11 +841,21 @@ fn do_xargs(args: &[&str]) -> Result<CommandResult, XargsError> {
...
+            .action(ArgAction::SetTrue),
         )
+        .arg(
+            Arg::new(options::REPLACE)
+                .long(options::REPLACE)
+                .short('I')
+                .short_alias('i')
+                .num_args(0..=1)
+                .value_parser(clap::value_parser!(String))
+                .help("Replace R in INITIAL-ARGS with names read from standard input; if R is unspecified, assume {}"),
```
cakebaker, line 856: This line breaks formatting with `fmt`.
```suggestion
                .help(
                    "Replace R in INITIAL-ARGS with names read from standard input; \
                    if R is unspecified, assume {}",
                ),
```

src/xargs/mod.rs
sylvestre: How do you handle this with clap? New Arg?

cakebaker: Yes, a new `Arg` without long option.

sylvestre: ok
will do @cakebaker 
do you mind if I do that in a new PR ?
thanks

cakebaker: No, I don't mind :)

src/xargs/mod.rs
```diff
@@ -813,11 +841,24 @@ fn do_xargs(args: &[&str]) -> Result<CommandResult, XargsError> {
...
+            .long(options::VERBOSE)
+            .help("Be verbose")
+            .action(ArgAction::SetTrue),
+        )
+        .arg(
+            Arg::new(options::REPLACE)
+                .long(options::REPLACE)
+                .short('I')
+                .short_alias('i')
+                .num_args(0..=1)
```
cakebaker, line 854: I think you have to enable `require_equals`, otherwise it is not possible to figure out whether there is a value specified or not when using something like `xargs --replace echo {} bar`. And there is also a test missing to ensure that both `--replace` (using `{}` as default value) and `--replace=value` work.

The other changes look fine.

sylvestre: <3 
it is now "perfect", thanks

src/xargs/mod.rs
```diff
@@ -818,6 +847,25 @@ fn do_xargs(args: &[&str]) -> Result<CommandResult, XargsError> {
...
+                .help(
+                    "Replace R in INITIAL-ARGS with names read from standard input; \
+                    if R is unspecified, assume {}",
+                ),
+        )
+        .arg(
+            Arg::new(options::REPLACE_I)
+                .short('I')
+                .num_args(1)
+                .hide(true)
```
cakebaker, line 866: Why do you hide the option? GNU `xargs` lists it with `--help`.
```suggestion
                .help("same as --replace=R")
                .value_name("R")
```

src/xargs/mod.rs
```diff
@@ -818,6 +847,25 @@ fn do_xargs(args: &[&str]) -> Result<CommandResult, XargsError> {
                 .help("Be verbose")
                 .action(ArgAction::SetTrue),
         )
+        .arg(
+            Arg::new(options::REPLACE)
+                .long(options::REPLACE)
+                .short('i')
+                .num_args(0..=1)
+                .require_equals(true)
+                .value_parser(clap::value_parser!(String))
```
cakebaker, line 856: The help text mentions `R`.
```suggestion
                .value_parser(clap::value_parser!(String))
                .value_name("R")
```

tests/xargs_tests.rs
```diff
@@ -386,3 +426,65 @@ fn xargs_zero_lines() {
...
+        .assert()
+        .stdout(predicate::str::contains("foo bar foo"));
+
+    // Combine the two options to see which one wins
+    Command::cargo_bin("xargs")
+        .expect("found binary")
+        .args(["-I=_", "-i", "echo", "{} bar {}"])
+        .write_stdin("foo")
+        .assert()
+        .stdout(predicate::str::contains("foo bar foo"));
```
cakebaker, line 480: This only tests the case with `-i` winning. However, if you swap the positions of `-I=_` and `-i`, `-I=_` should win, which currently is not the case (see also my previous comment about `override`: https://github.com/uutils/findutils/pull/323/files#r1547994855).

tests/xargs_tests.rs
```diff
@@ -386,3 +426,65 @@ fn xargs_zero_lines() {
...
+
+    // Excepted to fail
```
cakebaker, line 482:
```suggestion
    // Expected to fail
```

sylvestre (commented):
i think we are good this time :)

cakebaker:
Good work :)

sylvestre:
many thanks for this terrific review :)

Issue #307, opened 2024-01-29
Implement `-newerXY`


sylvestre:
The upstream manpage states:

```
-newerXY reference
          Compares the timestamp of the current file with reference.   The
          reference  argument  is  normally the name of a file (and one of
          its timestamps is used for the comparison) but it may also be  a
          string  describing  an  absolute time.  X and Y are placeholders
          for other letters, and these letters select which time belonging
          to how reference is used for the comparison.

          a   The access time of the file reference
          B   The birth time of the file reference
          c   The inode status change time of reference
          m   The modification time of the file reference
          t   reference is interpreted directly as a time
```

```
$ find . ! -newermt 'jan 01, 2000' -exec touch -d@1706517278 {} +   
Error: Unrecognized flag: '-newermt'
```

hanbings:
hi, it looks like there hasn't been an update here for a long time, can I give it a try?

cakebaker:
@hanbings sure, go ahead :)

hanbings:
Thanks. 

Now `find . ! -newermt 'jan 01, 2000' -exec touch -d@1706517278 {} + ` works, and `-newerat` and `-newerct` is also available.
`-newerBt` is available, but depends on whether the file system supports `birthed time / created time`.

Use the following command to better observe the results.
``` bash
find . -newermt 'jan 01, 2000'
find . -newermt 'jan 01, 2000 00:00:00'
```

In order to parse this time string, I added a function `parse_date_str_to_timestamps()` that converts from string to timestamp, then created the corresponding `Matcher` and compared it with the timestamp of the target file to get the result. I will add relevant test code later.

Please forgive my code for not looking very good.

PR #342, opened 2024-03-27
find: Implement `-newerXY`

hanbings:
link: https://github.com/uutils/findutils/issues/307

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -570,13 +570,11 @@ fn build_matcher_tree(
             }
 
             _ => {
-                if i >= args.len() - 1 {
-                    return Err(From::from(format!("missing argument to {}", args[i])));
-                }
                 match parse_str_to_newer_args(args[i]) {
                     Some((x_option, y_option)) => {
-                        // TODO Some code is also needed to bs compatible with different file
-                        // systems for file create time.
+                        if i >= args.len() - 1 {
+                            return Err(From::from(format!("missing argument to {}", args[i])));
+                        }
                         #[cfg(target_os = "linux")]
                         if x_option == "B" {
                             return Err(From::from("find: This system does not provide a way to find the birth time of a file."));
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -49,7 +49,10 @@ use self::quit::QuitMatcher;
 use self::regex::RegexMatcher;
 use self::size::SizeMatcher;
 use self::stat::{InodeMatcher, LinksMatcher};
-use self::time::{FileTimeMatcher, FileTimeType, NewerMatcher, NewerOptionType, NewerTimeMatcher};
+use self::time::{
+    FileTimeMatcher, FileTimeType, NewerMatcher, NewerOptionMatcher, NewerOptionType,
+    NewerTimeMatcher,
+};
 use self::type_matcher::TypeMatcher;
 
 use super::{Config, Dependencies};
@@ -290,7 +293,7 @@ fn parse_date_str_to_timestamps(date_str: &str) -> Option<i64> {
 /// X and Y are constrained to a/B/c/m and t.
 /// such as: "-neweraB" -> Some(a, B) "-neweraD" -> None
 fn parse_str_to_newer_args(input: &str) -> Option<(String, String)> {
-    let re = Regex::new(r"-newer([aBcmt])([aBcmt])").unwrap();
+    let re = Regex::new(r"-newer([aBcm])([aBcmt])").unwrap();
     if let Some(captures) = re.captures(input) {
         let x = captures.get(1)?.as_str().to_string();
         let y = captures.get(2)?.as_str().to_string();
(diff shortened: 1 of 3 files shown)


sylvestre:
Could you please add unit and integration tests? Thanks

hanbings:
Okay, I will add the test code as soon as possible.

sylvestre:
Thanks

cakebaker:
@hanbings apart from the tests, is the code ready for review?

hanbings:
> @hanbings apart from the tests, is the code ready for review?

I'm sorry I just saw the message. Yes, I've added some test code, but there seems to be something wrong with the Github Actions being tested, and I'm not quite sure the test code meets the coverage requirements. In addition, there is another piece of file system compatibility code that is not yet ready. But yeah, ready to review. Thank you.

src/find/matchers/mod.rs
```diff
@@ -348,6 +393,33 @@ fn build_matcher_tree(
                 i += 1;
                 Some(NewerMatcher::new(args[i])?.into_box())
             }
+            "-newerat" | "-newerBt" | "-newerct" | "-newermt" => {
```
cakebaker, line 396: This covers "only" a part of the possible combinations. From the man page:
```
-newerXY reference

Succeeds  if  timestamp  X  of the file being considered is newer than timestamp Y of the file reference.
The letters X and Y can be any of the following letters:

a   The access time of the file reference
B   The birth time of the file reference
c   The inode status change time of reference
m   The modification time of the file reference
t   reference is interpreted directly as a time

Some combinations are invalid; for example, it is invalid for X to be t. [...]
```
This means you can also have combinations like `-neweraa`, `-neweraB`, `-newerac`, `-neweram`, etc.

hanbings: Sorry, I misunderstood the documentation. I'll rewrite the code.

hanbings: I have completed the implementation of the parameter function described in the documentation and added a portion of the test code. Should we start the review?

src/find/mod.rs
cakebaker, line 1: The test function `find_newer_xy` currently tests five different things, so I suggest to break it up into five test functions to make it easier to read and understand.

hanbings: OK, I've committed the changes. Thanks.

src/find/mod.rs
```diff
@@ -854,4 +854,112 @@ mod tests {
...
+        let args = ["-newerat", "-newerBt", "-newerct", "-newermt"];
+        let times = [""];
+
+        for (arg, time) in args.iter().zip(times.iter()) {
+            let deps = FakeDependencies::new();
+            let rc = find_main(&["find", "./test_data/simple/subdir", arg, time], &deps);
+
+            assert_eq!(rc, 1);
+        }
+    }
```
cakebaker, line 964: It's good that you test for an empty time. Unfortunately, the tested behavior differs from the behavior of GNU `find`:
```
$ find test_data/simple/subdir -newerat ""
$ echo $?
0

$ cargo run --bin find test_data/simple/subdir/ -newerat ""
Error: find: I cannot figure out how to interpret ‘’ as a date or time
$ echo $?
1
```

src/find/mod.rs
```diff
@@ -854,4 +854,112 @@ mod tests {
...
+
+    #[test]
+    fn test_find_newer_xy_error_path() {
+        // Catch a parsing error.
+        #[cfg(target_os = "linux")]
+        let args = ["-newerat", "-newerct", "-newermt"];
+        #[cfg(not(target_os = "linux"))]
+        let args = ["-newerat", "-newerBt", "-newerct", "-newermt"];
+        let times = [""];
```
cakebaker, line 956: You don't need an array for a single value. Using `let time = "";` allows you to simplify the `for` loop below.

src/find/mod.rs
hanbings: Excuse me, I made some modifications to the code and this feature should have been implemented. 
In addition, I added a new function test_find_newer_xy_empty_time_parameter() specifically for testing empty time string input to solve the problem of time inconsistency between the file used by the test and the local environment file (to fix submissions where CI test errors occurred earlier).

src/find/matchers/mod.rs
```diff
@@ -252,6 +256,53 @@ fn convert_arg_to_comparable_value_and_suffix(
...
 }
 
+/// This is a function that converts a specific string format into a timestamp.
+/// It allows converting a time string of
+/// "(week abbreviation) (date), (year) (time)" to a Unix timestamp.
+/// such as: "jan 01, 2025 00:00:01" -> 1735689601000
+/// When (time) is not provided, it will be automatically filled in as 00:00:00
+/// such as: "jan 01, 2025" = "jan 01, 2025 00:00:00" -> 1735689600000
+fn parse_date_str_to_timestamps(date_str: &str) -> Option<i64> {
+    let regex_pattern = r"^(?:(\w{3} \d{2}))?(?:, (\d{4}))?(?: (\d{2}:\d{2}:\d{2}))?$";
```
sylvestre, line 266: please use this kind of patterns:
make it easier to understand 

    let regex_pattern = r#"^(?P<month_day>\w{3} \d{2})?(?:, (?P<year>\d{4}))?(?: (?P<time>\d{2}:\d{2}:\d{2}))?$"#;

src/find/matchers/time.rs
```diff
@@ -362,4 +504,155 @@ mod tests {
...
+        let file_info = get_dir_entry_for(&temp_dir.path().to_string_lossy(), "foo");
+        assert!(
+            created_matcher.matches(&file_info, &mut deps.new_matcher_io()),
+            "file created time should after 'time'"
+        );
+
+        // accessed time test
+        let accessed_matcher = NewerTimeMatcher::new(NewerOptionType::Accessed, time);
+
+        thread::sleep(Duration::from_secs(2));
```
sylvestre, line 594: why sleep ?

src/find/matchers/mod.rs
hanbings: Okay, this way of writing looks clearer! I will change it.

src/find/matchers/time.rs
hanbings: This is my oversight, sleep() is not really needed here, and there are two other codes with the same problem. They have been modified in the commit. Thanks!

src/find/matchers/time.rs
```diff
@@ -362,4 +504,149 @@ mod tests {
...
+            .now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap()
+            .as_millis()
+            .try_into()
+            .unwrap();
+
+        let created_matcher = NewerTimeMatcher::new(NewerOptionType::Birthed, time);
+
+        thread::sleep(Duration::from_secs(2));
```
sylvestre, line 575: same, is this sleep needed?

hanbings: This sleep() is used to ensure that there is a small distance between the time when assert() is executed and the time when the file was created. In fact, this code fails the test on my machine after removing the sleep().

and...
Currently the i64 type microsecond timestamp is used to determine the time. I have noticed that it is not accurate enough and I am going to change it to nanoseconds.

sylvestre: two seconds is long, try 0.1 ?

hanbings: `Duration::from_secs()` does not accept floating point arguments, maybe `Duration::from_millis(100)` would also be suitable?

PR #348, opened 2024-04-03
Run clippy pedantic fixes

sylvestre:
Done with:
$ cargo +nightly clippy --tests --fix --allow-dirty -- -W clippy::pedantic

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -74,6 +74,7 @@ impl<'a> MatcherIO<'a> {
         self.should_skip_dir = true;
     }
 
+    #[must_use]
     pub fn should_skip_current_dir(&self) -> bool {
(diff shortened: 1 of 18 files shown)


src/find/matchers/mod.rs
```diff
@@ -546,8 +549,8 @@ mod tests {
     use crate::find::tests::FakeDependencies;
     use walkdir::WalkDir;
 
-    /// Helper function for tests to get a DirEntry object. directory should
-    /// probably be a string starting with "test_data/" (cargo's tests run with
+    /// Helper function for tests to get a `DirEntry` object. directory should
+    /// probably be a string starting with "`test_data`/" (cargo's tests run with
```
cakebaker, line 553: This change looks incorrect, though I'm not sure what's correct...

PR #350, opened 2024-04-03
use cargo-dist + prepare 0.5.0

sylvestre:


sylvestre:
auto approving as it is generated code with:
https://opensource.axo.dev/cargo-dist/book/way-too-quickstart.html

Issue #25, opened 2017-06-14
Panic when piped into head


tavianator:
```
$ find | head
.
./.git
./.git/branches
./.git/hooks
./.git/hooks/applypatch-msg.sample
./.git/hooks/commit-msg.sample
./.git/hooks/post-update.sample
./.git/hooks/pre-applypatch.sample
./.git/hooks/pre-commit.sample
./.git/hooks/prepare-commit-msg.sample
$ RUST_BACKTRACE=1 ./target/debug/find | head
.
./.git
./.git/branches
./.git/hooks
./.git/hooks/applypatch-msg.sample
./.git/hooks/commit-msg.sample
./.git/hooks/post-update.sample
./.git/hooks/pre-applypatch.sample
./.git/hooks/pre-commit.sample
./.git/hooks/prepare-commit-msg.sample
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Os { code: 32, message: "Broken pipe" } }', src/libcore/result.rs:859
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::_print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::begin_panic
   6: std::panicking::begin_panic_fmt
   7: rust_begin_unwind
   8: core::panicking::panic_fmt
   9: core::result::unwrap_failed
             at /build/rust/src/rustc-1.17.0-src/src/libcore/macros.rs:29
  10: <core::result::Result<T, E>>::unwrap
             at /build/rust/src/rustc-1.17.0-src/src/libcore/result.rs:737
  11: <findutils::find::matchers::printer::Printer as findutils::find::matchers::Matcher>::matches
             at ./src/find/matchers/printer.rs:26
  12: <findutils::find::matchers::logical_matchers::AndMatcher as findutils::find::matchers::Matcher>::matches::{{closure}}
             at ./src/find/matchers/logical_matchers.rs:39
  13: <core::slice::Iter<'a, T> as core::iter::iterator::Iterator>::all::{{closure}}
             at /build/rust/src/rustc-1.17.0-src/src/libcore/slice.rs:1027
  14: <core::slice::Iter<'a, T>>::search_while
             at /build/rust/src/rustc-1.17.0-src/src/libcore/slice.rs:1118
[...]

PR #351, opened 2024-04-04
find: Fix panic when piped into head

hanbings:
fix: https://github.com/uutils/findutils/issues/25


cakebaker:
Thanks for your PR :)

Issue #27, opened 2017-06-14
Exit code unchanged when an error is encountered


tavianator:
E.g.

```
$ mkdir foo
$ chmod -rwx foo
$ ./target/debug/find foo
foo
Error: foo: permission denied
$ echo $?
0
$ find foo
foo
find: ‘foo’: Permission denied
$ echo $?
1
```

hanbings:
We handle errors caused by reading files in `process_dir()`, which conforms to the processing logic of GNU find. But internal errors that would cause `process_dir()` to be handled cannot be passed to the outer layer. Therefore adding a boolean error flag is used to pass a signal to the upper function to change the exit code when `process_dir()` has any error.

It might solve the problem, even if it's not very elegant :(

PR #352, opened 2024-04-06
find: Changed exit code when an error is encountered in `process_dir()`.

hanbings:
fix: https://github.com/uutils/findutils/issues/27

### DIFF
--- a/src/find/mod.rs
+++ b/src/find/mod.rs
@@ -134,6 +134,7 @@ fn process_dir<'a>(
     deps: &'a dyn Dependencies<'a>,
     matcher: &dyn matchers::Matcher,
     quit: &mut bool,
+    has_some_process_error: &mut bool,
 ) -> u64 {
     let mut found_count: u64 = 0;
     let mut walkdir = WalkDir::new(dir)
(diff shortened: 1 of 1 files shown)


cakebaker:
Can you please add a test to ensure we don't regress in the future?

hanbings:
> Can you please add a test to ensure we don't regress in the future?

OK, I've added a test for this fix.

src/find/mod.rs
```diff
@@ -149,10 +150,14 @@ fn process_dir<'a>(
     // WalkDirIterator::skip_current_dir for explanation.
     let mut it = walkdir.into_iter();
     while let Some(result) = it.next() {
+        let mut matcher_io = matchers::MatcherIO::new(deps);
```
cakebaker, line 153: What's the reason to move this here from the `Ok` case?

hanbings: Sorry, I forgot to move it back into `Ok()`. 
My initial writing method was to write the error flag in `matcher_io` (which means `matcher_io` needs to be used in `Err`). Later, I found that it was designed as an immutable structure and then passed it as a parameter to the `process_dir()` function.

hanbings: Changes have been committed in the latest. Thank you.

src/find/mod.rs
cakebaker, line 1: I agree with your [comment](https://github.com/uutils/findutils/issues/27#issuecomment-2041001313) that using `has_some_process_error` is not a very elegant solution :)

I think you can eliminate the variable by using `set_exit_code` from `uucore::error` instead. It's what we use in coreutils in similar situations.

src/find/mod.rs
```diff
@@ -182,18 +187,25 @@ fn do_find<'a>(args: &[&str], deps: &'a dyn Dependencies<'a>) -> Result<u64, Box
...
+            &mut has_some_process_error,
         );
         if quit {
             break;
         }
     }
+
+    if has_some_process_error {
+        return Err("There are some process errors.".into());
+    }
```
cakebaker, line 207: GNU `find` doesn't show an error after the processing, it only shows the errors you show on line 157.

src/find/mod.rs
hanbings: Looks better! Code has been reverted and uses `uucore::error`. 
to be honest, I didn't know about this function before.

src/find/mod.rs
hanbings: This code has been removed in the new commit. Thanks.

cakebaker:
Good work, thanks :)

Issue #29, opened 2017-06-14
Empty brackets accepted


tavianator:
I assume this isn't an intentional extension:

```
$ ./target/debug/find \( \)
.
./.git
...
```

mcharsley:
I worked on the basis "when in doubt, do what GNU's find does". I grant you that it's fairly useless behaviour, but 
 - it matches GNU find
 - it's harmless enough

tavianator:
But

```
$ find \( \)
find: invalid expression; empty parentheses are not allowed.
```

Anyway, I agree it's relatively harmless to allow it.

mcharsley:
My mistake. I tried

find /tmp \(\)

and killed the process before I saw "find: `()': No such file or directory"

putting a space between the brackets does indeed give an error. That should
be simple enough to fix.

thanks for the bug reports by the way. Nice to see someone's looking at the
code :-).

hanbings:
Hello, I tried to solve it by adding a special judgment to the Matcher of uutils/findutils, and it seems to work. I've submitted a PR. :)

PR #353, opened 2024-04-11
find: Fix unexpectedly accepted empty expression.

hanbings:
fix: https://github.com/uutils/findutils/issues/29

This commit modifies a test function `build_top_level_matcher_too_many_brackets` outside of my code, 
which tests that `find -true \( \) \)` will return a `too many ')'` error.

But in GNU find, this instruction will first cause an `empty parentheses are not allowed` error, 
so it is changed to `find . -type f \( -name "*.txt" \) \)` to maintain compatibility.

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -531,6 +531,14 @@ fn build_matcher_tree(
                 if !expecting_bracket {
                     return Err(From::from("you have too many ')'"));
                 }
+
+                let bracket = args[i - 1];
+                if bracket == "(" {
+                    return Err(From::from(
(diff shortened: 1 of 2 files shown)


sylvestre:
could you please add a test here: https://github.com/uutils/findutils/blob/main/tests/find_cmd_tests.rs
thanks

hanbings:
> could you please add a test here: https://github.com/uutils/findutils/blob/main/tests/find_cmd_tests.rs thanks

OK, it's included in the latest commit.

cakebaker:
Thanks!

Issue #221, opened 2023-03-28
Feature request: support -amin -cmin -mmin age ranges


xuyang2:
https://www.gnu.org/software/findutils/manual/html_node/find_html/Age-Ranges.html#Age-Ranges

hanbings:
Excuse me, Is anyone working on this? Can I try it?

cakebaker:
@hanbings I don't think anyone is working on it, so go ahead :)

PR #355, opened 2024-04-19
find: Implement `-amin` `-cmin` `-mmin` age ranges.

hanbings:
implement: https://github.com/uutils/findutils/issues/221

### DIFF
--- a/src/find/matchers/time.rs
+++ b/src/find/matchers/time.rs
@@ -695,4 +695,68 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn file_age_range_matcher() {
+        let file = get_dir_entry_for("test_data", "simple");
+
+        // mores test
+        // mocks:
+        // - find test_data/simple -mmin +1
+        // - find test_data/simple -amin +1
+        // Means to find files accessed / modified more than 1 minute ago.
+        [FileTimeType::Modified, FileTimeType::Accessed]
+            .iter()
+            .for_each(|time_type| {
+                let more_matcher =
+                    FileAgeRangeMatcher::new(*time_type, ComparableValue::MoreThan(1));
+                assert!(
+                    more_matcher.matches(&file, &mut FakeDependencies::new().new_matcher_io()),
+                    "more minutes old file should match more than 1 minute old."
+                );
+            });
+
+        // lesses test
+        // mocks:
+        // - find test_data/simple -mmin -1
+        // - find test_data/simple -amin -1
+        // Means to find files accessed / modified less than 1 minute ago.
+        [FileTimeType::Modified, FileTimeType::Accessed]
+            .iter()
+            .for_each(|time_type| {
+                let less_matcher =
+                    FileAgeRangeMatcher::new(*time_type, ComparableValue::LessThan(1));
+                assert!(
+                    !less_matcher.matches(&file, &mut FakeDependencies::new().new_matcher_io()),
+                    "less minutes old file should not match less than 1 minute old."
+                );
+            });
+
+        let temp_dir = Builder::new().prefix("example").tempdir().unwrap();
+        let temp_dir_path = temp_dir.path().to_string_lossy();
+        let new_file_name = "newFile";
+        // this has just been created, so should be newer
+        File::create(temp_dir.path().join(new_file_name)).expect("create temp file");
+        let new_file = get_dir_entry_for(&temp_dir_path, new_file_name);
+
+        // mocks:
+        // - find test_data/simple -cmin -1
+        // Means to find the new file created less than 1 minute.
+        let less_matcher =
+            FileAgeRangeMatcher::new(FileTimeType::Created, ComparableValue::LessThan(1));
+        assert!(
+            less_matcher.matches(&new_file, &mut FakeDependencies::new().new_matcher_io()),
+            "The file just created should be included in the less 1 minute match."
+        );
+
+        // mocks:
+        // - find test_data/simple -cmin +1
+        // Means to find the new file created more than 1 minute.
+        let more_matcher =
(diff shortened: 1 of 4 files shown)


src/find/matchers/time.rs
```diff
@@ -647,4 +695,79 @@ mod tests {
...
+    #[test]
+    fn file_age_range_matcher() {
+        let temp_dir = Builder::new().prefix("example").tempdir().unwrap();
+        let temp_dir_path = temp_dir.path().to_string_lossy();
+        let new_file_name = "newFile";
+        // this has just been created, so should be newer
+        File::create(temp_dir.path().join(new_file_name)).expect("create temp file");
+        let new_file = get_dir_entry_for(&temp_dir_path, new_file_name);
+
+        // mores test
```
sylvestre, line 708:
```suggestion
        // more test
```

hanbings: Thanks. Changed in new commit.

sylvestre:
please add integrations tests in https://github.com/uutils/findutils/blob/main/tests/find_cmd_tests.rs
and please tests with invalid values (like string for numerical values)

hanbings:
> please add integrations tests in https://github.com/uutils/findutils/blob/main/tests/find_cmd_tests.rs and please tests with invalid values (like string for numerical values)

I've added tests for unusual character type numbers. 
But I found that it is able to recognize error parameters such as `+-321` and `1%2`, which is different from the GNU find results. 

This has to do with the way the `convert_arg_to_comparable_value` function recognizes and converts numbers, meaning that this problem occurs with both the `a/c/mtime` and `a/c/mmin` arguments. Maybe I should fix them in a new PR?


```shell
$ find . -atime 1%2
find: invalid argument `1%2' to `-atime'
$ find . -amin 1%2
find: invalid argument `1%2' to `-amin'
$ ./target/debug/find . -atime 1%2
$ ./target/debug/find . -amin 1%2
./.git
./.git/objects/b1
./.git/objects/7f
./.git/objects/a8
./target/debug/deps
./target/debug/incremental/find_cmd_tests-8yeq2sf4g2ts
...
```

tests/find_cmd_tests.rs
```diff
@@ -518,3 +518,34 @@ fn expression_empty_parentheses() {
...
+
+#[test]
+fn find_age_range() {
+    let args = ["-amin", "-cmin", "-mmin"];
+    let times = ["-60", "-120", "-240", "+60", "+120", "+240"];
+    let time_strings = [
+        "\"-60\"", "\"-120\"", "\"-240\"", "\"-60\"", "\"-120\"", "\"-240\"",
+    ];
+
+    for (arg, time) in args.iter().zip(times.iter()) {
```
cakebaker, line 530: I think this piece of code doesn't do what you think it does ;-) It doesn't create the Cartesian product of `args` and `times`:
```
for (arg, time) in args.iter().zip(times.iter()) {
    println!("{arg}: {time}");
}
```
outputs:
```
-amin: -60
-cmin: -120
-mmin: -240
```

tests/find_cmd_tests.rs
```diff
@@ -518,3 +518,34 @@ fn expression_empty_parentheses() {
...
+        Command::cargo_bin("find")
+            .expect("the time should match")
+            .args(["test_data/simple", arg, time])
+            .assert()
+            .success()
+            .code(0);
+    }
+
+    for (arg, time_string) in args.iter().zip(time_strings.iter()) {
```
cakebaker, line 539: See my previous comment.

tests/find_cmd_tests.rs
hanbings: Agree. I replaced it with a nested loop. : )

tests/find_cmd_tests.rs
```diff
@@ -518,3 +518,38 @@ fn expression_empty_parentheses() {
...
+                .expect("the time should match")
+                .args(["test_data/simple", arg, time])
+                .assert()
+                .success()
+                .code(0);
+        }
+    }
+
+    for arg in args.iter() {
+        for time_string in time_strings.iter() {
```
cakebaker, line 542:
```suggestion
    for arg in args {
        for time_string in time_strings {
```

tests/find_cmd_tests.rs
```diff
@@ -518,3 +518,38 @@ fn expression_empty_parentheses() {
...
+            Command::cargo_bin("find")
+                .expect("the except time should not match")
```
cakebaker, line 544: I'm not sure about this one:
```suggestion
                .expect("the time should not match")
```

src/find/mod.rs
```diff
@@ -736,6 +736,29 @@ mod tests {
...
+    // and The files on the local branch may be several days old.
+    //
+    // So this test may not be too accurate and can only ensure that
+    // the function can be correctly identified.
+    #[test]
+    fn find_amin_cmin_mmin() {
+        let args = ["-amin", "-cmin", "-mmin"];
+        let times = ["-60", "-120", "-240", "+60", "+120", "+240"];
+
+        for (arg, time) in args.iter().zip(times.iter()) {
```
cakebaker, line 754: Here you need a nested loop to get the expected result because `zip` doesn't generate the Cartesian product of `args` and `times`.

hanbings:
Thanks, the conflict has been resolved and the code modified in new commits.

sylvestre:
it would be nice to work on https://github.com/uutils/findutils/issues/360 next
it will be easier to detect improvement / regression wrt the GNU test suite

hanbings:
> it would be nice to work on #360 next it will be easier to detect improvement / regression wrt the GNU test suite

Thanks! I will complete it as soon as I can.

PR #357, opened 2024-04-22
clippy: Fix clippy warning.

hanbings:
I'm sorry that I wrote some redundant code at https://github.com/uutils/findutils/pull/353 and https://github.com/uutils/findutils/pull/342 which caused `cargo clippy --all-targets --all-features` to generate multiple warnings.

Fix them in this PR.


cakebaker:
No need to be sorry :)  I added `--all-targets` to the CI in https://github.com/uutils/findutils/pull/359, so in the future the CI should catch such issues.

Thanks for fixing the issues!

PR #361, opened 2024-04-23
find: Fix `convert_arg_to_comparable_value()` function parsing unexpected characters.

hanbings:
convert_arg_to_comparable_value() as mentioned in https://github.com/uutils/findutils/pull/355#issuecomment-2067613423 parses some strange characters. 

The original regular expression was `([+-]?)(\d+)$` , which did a partial match and only included `1%2` of `2` as part of the match. 
After adding the `^` limiter (`^([+-]?)(\d+)$`), the expression can be restricted to only match `+` `-` or `0-9`.

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -220,7 +220,7 @@ fn convert_arg_to_comparable_value(
     option_name: &str,
     value_as_string: &str,
 ) -> Result<ComparableValue, Box<dyn Error>> {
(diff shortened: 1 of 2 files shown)


hanbings:
Thanks. 
I added an overall test for the -a/c/mtime directive in find_cmd_tests.rs. 

But it may be that Codecov does not detect the diff because the PR only fixes regular expressions, so the coverage changes cannot be seen in the Codecov report.

PR #363, opened 2024-04-28
xargs: do not merge extra args before replacing

### LINKED ISSUE #362: xargs: stdin should not be merged before replacing
Given the following stdin,

```
abc
def
g
```

`xargs -i echo [{}]` should give

```
[abc]
[def]
[g]
```

instead of `[abc def g]`.

<summary>Test using python</summary>

(Ubuntu)

```python
>>> from subprocess import run

# xargs 0.5.0 from uutils
>>> run(['./xargs', '-i', 'echo', '[{}]'], capture_output=True, input=b"abc\ndef\ng").stdout
b'[abc def g]\n'

# xargs 4.8.0 from GNU
>>> run(['/usr/bin/xargs', '-i', 'echo', '[{}]'], capture_output=True, input=b"abc\ndef\ng").stdout
b'[abc]\n[def]\n[g]\n'
```

<summary>Test using shell</summary>

(MSYS2 on Windows)

```shell
# xargs 0.5.0 from uutils
$ printf 'abc\ndef\ng' | ./xargs -i echo [{}]
[abc def g]

# xargs 4.9.0 from GNU
$ printf 'abc\ndef\ng' | xargs -i echo [{}]
[abc]
[def]
[g]
```

**Platform**: unknown-linux-gnu, pc-windows-msvc, other not tested.

## Relavant codes

https://github.com/uutils/findutils/blob/baa09baf47992a074538ed0e4bb3c4e952f37111/src/xargs/mod.rs#L407-L435

There's no test on this behaviour.

https://github.com/uutils/findutils/blob/baa09baf47992a074538ed0e4bb3c4e952f37111/tests/xargs_tests.rs#L429-L430

Relates-to: #323

YDX-2147483647:
Resolves #362

- If `-I`/`-i`/`--replace` is given, now we process extra args (read from stdin) _one by one_, and replace all matches in initial args (given in argv) with the same extra arg in each time. (i.e. Cartesian product)

```shell
# Before
$ printf 'a    bc\ndef\ng' | xargs -I _ echo '[_]'
[a bc def g]

# After
$ printf 'a    bc\ndef\ng' | cargo run --bin xargs -- -I _ echo '[_]'
[a    bc]
[def]
[g]
```

- `-L` (`--max-lines`), `-n` (`--max-args`) and `-I`/`-i` are mutually exclusive. Now we warn if more than one are given.

### DIFF
--- a/src/xargs/mod.rs
+++ b/src/xargs/mod.rs
@@ -406,14 +406,9 @@ impl CommandBuilder<'_> {
         let mut command = Command::new(entry_point);
 
         if let Some(replace_str) = &self.options.replace {
-            // we replace the first instance of the replacement string with
-            // the extra args, and then replace all instances of the replacement
-            let replacement = self
-                .extra_args
-                .iter()
-                .map(|s| s.to_string_lossy())
-                .collect::<Vec<_>>()
-                .join(" ");
+            // Replace all occurrences in initial args with the extra arg,
+            // Thanks to `MaxArgsCommandSizeLimiter`, we only process a single extra arg here.
+            let replacement = self.extra_args[0].to_string_lossy();
             let initial_args: Vec<OsString> = initial_args
                 .iter()
                 .map(|arg| {
@@ -797,7 +792,7 @@ fn do_xargs(args: &[&str]) -> Result<CommandResult, XargsError> {
                 .long(options::MAX_ARGS)
                 .help(
                     "Set the max number of arguments read from stdin to be passed \
-                    to each command invocation (mutually exclusive with -L)",
+                    to each command invocation (mutually exclusive with -L and -I/-i)",
                 )
                 .value_parser(validate_positive_usize),
         )
@@ -807,7 +802,7 @@ fn do_xargs(args: &[&str]) -> Result<CommandResult, XargsError> {
                 .long(options::MAX_LINES)
                 .help(
                     "Set the max number of lines from stdin to be passed to each \
-                    command invocation (mutually exclusive with -n)",
+                    command invocation (mutually exclusive with -n and -I/-i)",
                 )
                 .value_parser(validate_positive_usize),
         )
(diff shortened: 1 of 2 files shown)


YDX-2147483647:
Hi! I've fixed the issue in normal cases, but if conflict options are given, things get complicated.

Currently, we throw a warning and take the last option.

https://github.com/uutils/findutils/blob/baa09baf47992a074538ed0e4bb3c4e952f37111/src/xargs/mod.rs#L930-L933

Should I follow this behaviour and [change 100+ lines](https://github.com/uutils/findutils/pull/363/commits/31a9c11dc27e919b5e6650fb7888cf83b1427fba)? Or may I use [`Arg::conflicts_with` in clap](https://docs.rs/clap/4.5.4/clap/struct.Arg.html#method.conflicts_with) and throw an error?

(Modified from https://github.com/uutils/findutils/issues/362#issuecomment-2081512700 )

YDX-2147483647:
@sylvestre Hello, may I ask your opion on https://github.com/uutils/findutils/pull/363#issuecomment-2081766265? (No need to review codes, just a design question.)

sylvestre:
@YDX-2147483647 sorry, i missed your question :(
you should do the same as the GNU implementation

YDX-2147483647:
Never mind and thanks for your answer.

So those 100+ lines is worth changing. I'll add more tests when I have time.

YDX-2147483647:
Well, I've updated the branch. Could someone approve the CI? I think this PR is ready now.

sylvestre:
it it ready for review? :)
thanks

YDX-2147483647:
Yes! It is ready.

sylvestre:
nice:
`Warning: Changes from main: PASS +4 / FAIL -4 / ERROR +0 / SKIP +0 `

PR #368, opened 2024-05-07
find: Implement `-[no]user` and `-[no]group` predicates.

hanbings:
implement: https://github.com/uutils/findutils/issues/7

This PR implements `-user` and `-group`, ~~and also adds `-nouser` and `-nogroup` predicates to filter out files outside a specific user or group.~~ and add `-nouser` and `-nogroup` to filter the files of deleted users or deleted groups.

- Added dependency `nix = { version = "0.28", features = ["user"] }` for parsing users and groups to correctly convert to `uid` and `gid`.
- Not implemented for Windows platforms because `uid` and `gid` or equivalent metadata are not provided in `std::os::windows::fs::MetadataExt`.


hanbings:
**Reimplemented `-nouser` and `-nogroup`.**

Used to retrieve files for which user and group attributes do not exist (but uid or gid may exist).
For example:

- Copy the files from Windows to a USB drive, then plug in the Linux device and retrieve the files.
- Create a user on Linux, create files using the new user created, then delete the new user and retrieve the created files.

src/find/matchers/group.rs
```diff
@@ -0,0 +1,141 @@
+use super::Matcher;
```
sylvestre, line 1: Please add the license header

hanbings: I'm sorry I'm not sure which copyright holder to claim.
How about citing the project's License using the following text?

```
// This file is part of the uutils find package.
//
// For the full copyright and license information, please view the LICENSE
// file that was distributed with this source code.
```

sylvestre:
The CI fails, could you please have a look? Thanks

hanbings:
> The CI fails, could you please have a look? Thanks

Sorry for the late reply, I have submitted the fix related to the current PR.
There is also another issue with file paths in Windows tests, but it is not part of this. I will open a new PR to fix it later.

hanbings:
Hi, I have completed the content of https://github.com/uutils/findutils/issues/371 (`-uid` and `-gid`) on my branch, I think these features have the same characteristics as `-user` and `-group` to form a complete function and the amount of code is not very large (30 lines of functional code, the rest is test code). 

I would like to merge them into the branch of the current PR. How about it?

sylvestre:
With the GNU testsuite:
`Warning: Changes from main: PASS +0 / FAIL +0 / ERROR +0 / SKIP +0 `
BFS:
`Warning: Changes from main: PASS +4 / SKIP +0 / FAIL -4`

Nice :)
would be nice to finish
https://github.com/uutils/findutils/pull/400

PR #369, opened 2024-05-10
find: Implement `-ok[dir]` predicate.

### LINKED ISSUE #8: FeatureComplete: implement ok[dir] predicate
This should be a minor extension to the existing -exec[dir] functionality

hanbings:
implement: https://github.com/uutils/findutils/issues/8

### DIFF
--- a/src/find/matchers/exec.rs
+++ b/src/find/matchers/exec.rs
@@ -6,7 +6,7 @@
 
 use std::error::Error;
 use std::ffi::OsString;
-use std::io::{stderr, stdin, Write};
+use std::io::{stderr, stdin, stdout, Write};
 use std::path::Path;
 use std::process::Command;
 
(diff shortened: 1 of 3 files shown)


sylvestre:
@hanbings do you have an update on this ? thanks :)

src/find/matchers/exec.rs
```diff
@@ -83,6 +86,25 @@ impl Matcher for SingleExecMatcher {
...
+
+        // support interactive exec
+        if self.interactive {
+            let tips = format!(
+                "{} ... {} > ? [y/n]: ",
+                self.executable,
+                path_to_file.to_string_lossy()
+            );
+            #[allow(clippy::explicit_write)]
+            write!(stdout(), "{}", tips).unwrap();
```
tavianator, line 98: POSIX [specifies](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/find.html#tag_20_47_11)

- The -ok primary shall write a prompt to *standard error*
- In the POSIX locale, the last non-&lt;blank> in the prompt shall be '?'

hanbings: Thanks! I overlooked these, I will fix them later.

sylvestre:
@hanbings any plan to finish it ? thanks :)

hanbings:
> @hanbings any plan to finish it ? thanks :)

Yes, I’ll make sure to get them done, but the implementation is more challenging than it seems. I’ll need to free up some time, so I expect to be able to work on them in about one to two weeks. (same https://github.com/uutils/findutils/pull/411#issuecomment-2777762956)

Issue #384, opened 2024-05-12
Improve parsing of Date input formats


hanbings:
There are 11 time parsing formats mentioned in the GNU findutils manual: https://www.gnu.org/software/findutils/manual/html_node/find_html/Date-input-formats.html

Include:
[General date syntax](https://www.gnu.org/software/findutils/manual/html_node/find_html/General-date-syntax.html)
[Calendar date items](https://www.gnu.org/software/findutils/manual/html_node/find_html/Calendar-date-items.html)
[Time of day items](https://www.gnu.org/software/findutils/manual/html_node/find_html/Time-of-day-items.html)
[Time zone items](https://www.gnu.org/software/findutils/manual/html_node/find_html/Time-zone-items.html)
[Combined date and time of day items](https://www.gnu.org/software/findutils/manual/html_node/find_html/Combined-date-and-time-of-day-items.html)
[Day of week items](https://www.gnu.org/software/findutils/manual/html_node/find_html/Day-of-week-items.html)
[Relative items in date strings](https://www.gnu.org/software/findutils/manual/html_node/find_html/Relative-items-in-date-strings.html)
[Pure numbers in date strings](https://www.gnu.org/software/findutils/manual/html_node/find_html/Pure-numbers-in-date-strings.html)
[Seconds since the Epoch](https://www.gnu.org/software/findutils/manual/html_node/find_html/Seconds-since-the-Epoch.html)
[Specifying time zone rules](https://www.gnu.org/software/findutils/manual/html_node/find_html/Specifying-time-zone-rules.html)

However, we currently only support parsing of `General date syntax`.

tertsdiepraam:
This probably fits what We're doing in the `parse_datetime` repository!

hanbings:
cool! It seems so. 
I would use it as a dependency to replace the existing parsing function. : )

Issue #370, opened 2024-05-10
Implement `-anewer` and `-cnewer`


hanbings:
refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Comparing-Timestamps.html

```
Test: -anewer reference
Test: -cnewer reference
Test: -newer reference

True if the time of the last access (or status change or data modification) of the current file is more recent 
than that of the last datamodification of the reference file. 
As such, ‘-anewer’ is equivalent to ‘-neweram’, ‘-cnewer’ to ‘-newercm’, and ‘-newer’ to ‘-newermm’.
```

PR #386, opened 2024-05-18
find: Implement `-anewer` and `-cnewer`.

hanbings:
implement: https://github.com/uutils/findutils/issues/370

Added extra judgment to `parse_str_to_newer_args()` function.
1. `-anewer` is equivalent to `-neweram`
2. `-cnewer` is equivalent to `-newercm`
3. `-newer` is equicalent to `-newermm`

BFS test `+1 PASS` in commit 57f2a3bc0751f617933f551a531bdf45acdf77fd.

This PR **does not** solve [Parsing multiple time formats](https://github.com/uutils/findutils/issues/384), I will open a new PR to handle it separately. : )

### DIFF
--- a/tests/find_cmd_tests.rs
+++ b/tests/find_cmd_tests.rs
@@ -518,3 +518,50 @@ fn expression_empty_parentheses() {
         ))
         .stdout(predicate::str::is_empty());
 }
+
+#[test]
+#[serial(working_dir)]
+fn find_newer_xy() {
+    #[cfg(target_os = "linux")]
+    let x_options = ["a", "c", "m"];
+    #[cfg(not(target_os = "linux"))]
+    let x_options = ["a", "B", "c", "m"];
+    #[cfg(target_os = "linux")]
+    let y_options = ["a", "c", "m"];
+    #[cfg(not(target_os = "linux"))]
+    let y_options = ["a", "B", "c", "m"];
+
+    for &x in &x_options {
+        for &y in &y_options {
+            let arg = &format!("-newer{x}{y}").to_string();
+            Command::cargo_bin("find")
+                .expect("found binary")
+                .args([
+                    "./test_data/simple/subdir",
+                    arg,
+                    "./test_data/simple/subdir/ABBBC",
+                ])
+                .assert()
+                .success()
+                .stderr(predicate::str::is_empty());
+        }
+    }
+
+    #[cfg(target_os = "linux")]
+    let args = ["-newerat", "-newerct", "-newermt"];
+    #[cfg(not(target_os = "linux"))]
+    let args = ["-newerat", "-newerBt", "-newerct", "-newermt"];
+    let times = ["jan 01, 2000", "jan 01, 2000 00:00:00"];
+
+    for &arg in &args {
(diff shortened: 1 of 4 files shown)


tests/find_cmd_tests.rs
```diff
@@ -518,3 +518,50 @@ fn expression_empty_parentheses() {
...
+#[serial(working_dir)]
+fn find_newer_xy() {
+    #[cfg(target_os = "linux")]
+    let x_options = ["a", "c", "m"];
+    #[cfg(not(target_os = "linux"))]
+    let x_options = ["a", "B", "c", "m"];
+    #[cfg(target_os = "linux")]
+    let y_options = ["a", "c", "m"];
+    #[cfg(not(target_os = "linux"))]
+    let y_options = ["a", "B", "c", "m"];
```
cakebaker, line 532: As `x_options` and `y_options` contain the same values, I think it is unnecessary to have two vars. And so my suggestion is to simply use `options`.

tests/find_cmd_tests.rs
```diff
@@ -518,3 +518,50 @@ fn expression_empty_parentheses() {
...
+
+    for &x in &x_options {
+        for &y in &y_options {
+            let arg = &format!("-newer{x}{y}").to_string();
```
cakebaker, line 536: Calling `to_string()` is unnecessary because `format!` already returns a string.
```suggestion
            let arg = &format!("-newer{x}{y}");
```

tests/find_cmd_tests.rs
```diff
@@ -518,3 +518,50 @@ fn expression_empty_parentheses() {
...
+    let x_options = ["a", "c", "m"];
+    #[cfg(not(target_os = "linux"))]
+    let x_options = ["a", "B", "c", "m"];
+    #[cfg(target_os = "linux")]
+    let y_options = ["a", "c", "m"];
+    #[cfg(not(target_os = "linux"))]
+    let y_options = ["a", "B", "c", "m"];
+
+    for &x in &x_options {
+        for &y in &y_options {
```
cakebaker, line 535: I think the references are not necessary and you can simply use `for x in x_options`.

tests/find_cmd_tests.rs
```diff
@@ -518,3 +518,50 @@ fn expression_empty_parentheses() {
...
+
+    #[cfg(target_os = "linux")]
+    let args = ["-newerat", "-newerct", "-newermt"];
+    #[cfg(not(target_os = "linux"))]
+    let args = ["-newerat", "-newerBt", "-newerct", "-newermt"];
+    let times = ["jan 01, 2000", "jan 01, 2000 00:00:00"];
+
+    for &arg in &args {
+        for &time in &times {
+            let arg = &format!("{arg}{time}").to_string();
```
cakebaker, line 558: See my previous comments.

hanbings:
Thanks. I'm sorry for these issues, they have been fixed in new commits.

cakebaker:
Thanks. There's no need to be sorry, the things I mentioned are minor details ;-)  And it's great that you fixed them in `time.rs`, too.

Issue #373, opened 2024-05-11
Implement `-samefile`


hanbings:
refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Hard-Links.html

```
Test: -samefile NAME
True if the file is a hard link to the same inode as name. 
This implies that name and the file reside on the same file system, i.e., 
they have the same device number.

Unless the ‘-L’ option is also given to follow symbolic links, 
one may confine the search to one file system by using the ‘-xdev’ option. 
This is useful because hard links cannot point outside a single file system, 
so this can cut down on needless searching.

If the ‘-L’ option is in effect, then dereferencing of symbolic links applies 
both to the name argument of the ‘-samefile’ primary and to each file examined 
during the traversal of the directory hierarchy. Therefore, ‘find -L -samefile NAME’ 
will find both hard links and symbolic links pointing to the file referenced by name.
```

PR #389, opened 2024-05-24
find: Implement `-samefile`

hanbings:
implement: https://github.com/uutils/findutils/issues/373

BFS test `+1 PASS` and GNU test `PASS +8` / `FAIL -20` in commit bc8ae9859ef41b21a67ab0b72143751240659e3a

### DIFF
--- a/src/find/matchers/samefile.rs
+++ b/src/find/matchers/samefile.rs
@@ -18,17 +18,8 @@ impl SameFileMatcher {
 }
 
 impl Matcher for SameFileMatcher {
-    #[cfg(unix)]
     fn matches(&self, file_info: &walkdir::DirEntry, _matcher_io: &mut super::MatcherIO) -> bool {
         uucore::fs::paths_refer_to_same_file(file_info.path(), &self.path, true)
-    }
-
-    #[cfg(not(unix))]
-    fn matches(&self, _file_info: &walkdir::DirEntry, _matcher_io: &mut super::MatcherIO) -> bool {
-        // FIXME
-        // MetadataExt under Windows system does not have an interface for obtaining inode,
-        // so the implementation of this function needs to introduce other libraries or use unsafe code.
-        false
     }
 }
 
@@ -37,7 +28,6 @@ mod tests {
     use std::fs;
 
     #[test]
-    #[cfg(unix)]
     fn test_samefile() {
         use crate::find::{
             matchers::{samefile::SameFileMatcher, tests::get_dir_entry_for, Matcher},
--- a/src/find/matchers/samefile.rs
+++ b/src/find/matchers/samefile.rs
@@ -28,6 +28,39 @@ impl Matcher for SameFileMatcher {
 
     #[cfg(not(unix))]
     fn matches(&self, _file_info: &walkdir::DirEntry, _matcher_io: &mut super::MatcherIO) -> bool {
+        // FIXME
+        // MetadataExt under Windows system does not have an interface for obtaining inode,
+        // so the implementation of this function needs to introduce other libraries or use unsafe code.
         false
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::fs;
+
+    #[test]
+    #[cfg(unix)]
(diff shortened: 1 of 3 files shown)


tests/find_cmd_tests.rs
```diff
@@ -587,3 +587,54 @@ fn find_newer_xy() {
         }
     }
 }
+
+#[test]
+#[serial(working_dir)]
+#[cfg(target_os = "linux")]
```
tavianator, line 593:
```suggestion
#[cfg(unix)]
```

hanbings: Thanks. Modified in latest commit. This target is more suitable than `linux`. : )

src/find/matchers/samefile.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub fn new(metadata: Metadata) -> SameFileMatcher {
+        SameFileMatcher { metadata }
+    }
+}
+
+impl Matcher for SameFileMatcher {
+    #[cfg(unix)]
+    fn matches(&self, file_info: &walkdir::DirEntry, _matcher_io: &mut super::MatcherIO) -> bool {
+        let meta = file_info.metadata().unwrap();
+        meta.ino() == self.metadata.ino()
```
tavianator, line 26: You also need to check `meta.dev()`.  Otherwise you'll return true for two unrelated files on different filesystems that happen to get the same inode number.

tavianator (changes requested):

src/find/matchers/samefile.rs
hanbings: Okay, I've added a check.

sylvestre: we have 
https://github.com/uutils/coreutils/blob/8cac375dddb4d4e87519c137c45fa831ac7f2619/src/uucore/src/lib/features/fs.rs#L559
it would not be a good candidate ?

hanbings: > we have https://github.com/uutils/coreutils/blob/8cac375dddb4d4e87519c137c45fa831ac7f2619/src/uucore/src/lib/features/fs.rs#L559 it would not be a good candidate ?

Thanks! This has now been changed to `uucore::fs::paths_refer_to_same_file` in https://github.com/uutils/findutils/pull/389/commits/f612c0da06859beed4aeab498f6195eafe8de7b2 and it works great!

sylvestre:
And I waiting for https://github.com/uutils/findutils/pull/400 to review this first :)

sylvestre:
@hanbings is that expected ? :)
_PASS +8 / FAIL -20 / ERROR +0 / SKIP +0_

hanbings:
If we mean the GNU/BFS tests for the code in this PR itself, yes.

src/find/matchers/samefile.rs
```diff
@@ -0,0 +1,60 @@
...
+
+impl Matcher for SameFileMatcher {
+    #[cfg(unix)]
+    fn matches(&self, file_info: &walkdir::DirEntry, _matcher_io: &mut super::MatcherIO) -> bool {
+        uucore::fs::paths_refer_to_same_file(file_info.path(), &self.path, true)
+    }
+
+    #[cfg(not(unix))]
+    fn matches(&self, _file_info: &walkdir::DirEntry, _matcher_io: &mut super::MatcherIO) -> bool {
+        // FIXME
```
sylvestre, line 28: please create an issue for this once it landed

sylvestre: also, it supports windows?
https://github.com/uutils/coreutils/blob/92665144c9ddf100d5044dc0c52af122a94587d0/src/uu/split/src/platform/windows.rs#L47

hanbings: > also, it supports windows? https://github.com/uutils/coreutils/blob/92665144c9ddf100d5044dc0c52af122a94587d0/src/uu/split/src/platform/windows.rs#L47

Oh! That's right! Changed to support multiple platforms. Thanks!

PR #393, opened 2024-05-29
find: make `InodeMatcher` & `LinksMatcher` unix-only

cakebaker:
This PR is a refactoring that makes the implementations of `InodeMatcher` and `LinksMatcher` unix-only. And moves the error handling for other platforms to where we handle `-inum` and `-links`.

The refactoring is done to get rid of two "field x is never read" warnings on Windows (see, for example, https://github.com/uutils/findutils/actions/runs/9271426702/job/25506758917#step:6:185)

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -21,6 +21,7 @@ mod prune;
 mod quit;
 mod regex;
 mod size;
+#[cfg(unix)]
 mod stat;
 mod time;
 mod type_matcher;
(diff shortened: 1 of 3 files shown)


sylvestre:
@hanbings maybe we should display it only when here are changes.
it will be too noisy otherwise

hanbings:
> @hanbings maybe we should display it only when here are changes. it will be too noisy otherwise

OK, I'll open a new PR to work on it.

PR #394, opened 2024-05-29
find: Fix `-newerXY` test code for Windows platform.

hanbings:
1. Using `fs::remove_file` in Windows platform tests may return an `Ok()` even [if the file is not deleted immediately](https://doc.rust-lang.org/std/fs/fn.remove_file.html), which will cause an error in the code in the test that assumes the file is deleted.
2. In another test code, some `/` separators were accidentally escaped as `\\`. Now it has been changed to use `contains` to determine whether the output contains a specific path and ignore the path separator problem.


sylvestre:
please add tests to make sure we don't regress, thanks

cakebaker:
@sylvestre: hm, what tests are you looking for? As far as I see it, this PR is about fixing the broken tests on Windows, so the tests are already there.

sylvestre:
oups, sorry

PR #400, opened 2024-06-07
Provide GNU test comparison comments for PRs in Github Actions.

hanbings:
link: https://github.com/uutils/findutils/issues/360

Add the PR comment code to `compat.yml`.
But I don't have a good way to test this code yet, so I'm setting it to `Daft` for now.

Issue #360, opened 2024-04-23
Just like with rust/coreutils, compare with the GNU implementation and highlight improvements and regressions


sylvestre:
This is done here:
https://github.com/uutils/coreutils/blob/main/.github/workflows/GnuTests.yml
and
https://github.com/uutils/coreutils/blob/main/.github/workflows/GnuComment.yml

we are publishing results on:
https://github.com/uutils/coreutils-tracking/

Note that we also have:
https://github.com/uutils/findutils-tracking/

PR #404, opened 2024-06-23
Fix CI code for `comment.yml`

hanbings:
link: https://github.com/uutils/findutils/issues/360 and proceed: https://github.com/uutils/findutils/pull/400

### DIFF
--- a/.github/workflows/comment.yml
+++ b/.github/workflows/comment.yml
@@ -1,15 +1,17 @@
 on:
-    workflow_run:
-      workflows: [External testsuites]
-      types: [completed]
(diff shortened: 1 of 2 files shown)


hanbings:
I've tested this code with PRs in my fork. There should be no problem.
and... I seem to have forgotten to do a commit squash to clean up the extra commit records. Could you clean them up when you merge them? Thanks! :)

sylvestre:
https://github.com/uutils/findutils/pull/389#issuecomment-2185344037
seems that it works, bravo :)

PR #407, opened 2024-06-24
Comments are only sent when GNU/BFS test compatibility changes.

hanbings:
Tested in fork, this change may still only take effect after being merged.

### DIFF
--- a/.github/workflows/comment.yml
+++ b/.github/workflows/comment.yml
@@ -72,5 +72,5 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: annotations.pull_request_number,
-              body: 'Commit ' + github.event.workflow_run.head_sha + ' has GNU testsuite comparison:\n```\n' + annotationContent + '\n```\n'
+              body: 'Commit ' + ${{ github.event.workflow_run.head_sha }} + ' has GNU testsuite comparison:\n```\n' + annotationContent + '\n```\n'
             });
--- a/.github/workflows/comment.yml
+++ b/.github/workflows/comment.yml
@@ -59,8 +59,8 @@ jobs:
               .join('\n');
 
             // check if no changes
-            let gnuTestReport = annotationsContent.includes('Run GNU findutils tests: Changes from main: PASS +0 / FAIL +0 / ERROR +0 / SKIP +0');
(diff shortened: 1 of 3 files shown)


.github/workflows/comment.yml
```diff
@@ -58,7 +58,14 @@ jobs:
               .map(annotation => `${annotation.run}: ${annotation.annotation.message}`)
               .join('\n');
 
-            console.log(annotationContent);
+            // check if no changes
+            let gnuTestReport = annotationContent.includes('Run GNU findutils tests: Changes from main: PASS +0 / FAIL +0 / ERROR +0 / SKIP +0');
```
sylvestre, line 62: this seems a bit fragile. In the json, we don't have the number to do some if > 0 ?

hanbings: That might mean getting the results for the master branch in comment.yml and comparing them to bfs-result.json and gnu-result.json? I think this might be a bit repetitive and complicated.

How about changing compare_bfs_result.py to send a Github Actions Annotations using `print(f"::warning :: BFS tests no change.")`  and catch annotation in `comment.yml` when compatibility is unchanged?

sylvestre: yeah, probably better and easier

sylvestre:
i guess it isn't ready yet, right ? :)

hanbings:
Yeah, I haven't tested this code on my fork yet.
I originally wanted to mark this PR as Draft, but Github doesn't seem to show me that button. I'll mention you here when I'm done, thanks! :)

hanbings:
@sylvestre done. 
This is a sample generated during testing: https://github.com/hanbings/findutils/pull/19
Is there anything else we need to add?

sylvestre:
sorry, looks like i wasn't cleary. What I would like is to have nothing to display when
BFS tests or Gnu tests = No changes

we should only display something when we noticed changes in bfs or gnu tests

hanbings:
> sorry, looks like i wasn't cleary. What I would like is to have nothing to display when BFS tests or Gnu tests = No changes
> 
> we should only display something when we noticed changes in bfs or gnu tests

No, actually it does not send comments when it contains `No changes`. But Github Actions will only execute the workflows file of the default branch, and the logic of whether to send comments is in the `comment.yml` file, so we need to merge this PR first to see the final effect.

sylvestre:
ok, let's see :)

PR #409, opened 2024-06-27
samefile: Cache FileInformation for the target

tavianator:
And don't follow symlinks.


sylvestre:
@hanbings wdyt ? :)

hanbings (commented):
That looks good to me, thanks! :)

Issue #366, opened 2024-05-04
Fix: `Error: -exec [args...] + isn't supported yet. Only -exec [args...] ;`


sylvestre:
`Error: -exec [args...] + isn't supported yet. Only -exec [args...] ;`
example:
` find . ! -newermt 'jan 01, 2000' -exec touch -d@1714807765 {} +`

hanbings:
Looks like what is mentioned at https://github.com/uutils/findutils/issues/6. 
In addition to `exec`, it is also necessary to support the `+` syntax of `execdir`.

winstonma:
Yep I faced this issue after replacing findutils through [oxidizr](https://github.com/jnsgruk/oxidizr) in Ubuntu 24.10:

This is a snippet of the script that uses `find` in `/usr/share/initramfs-tools/hooks/udev`
```
...
find -L /usr/lib/systemd/network -name '*.link*' -execdir cp -r -pt "$DESTDIR/usr/lib/systemd/network/" '{}' +
...
```

hanbings:
> Yep I faced this issue after replacing findutils through [oxidizr](https://github.com/jnsgruk/oxidizr) in Ubuntu 24.10:
> 
> This is a snippet of the script that uses `find` in `/usr/share/initramfs-tools/hooks/udev`
> 
> ```
> ...
> find -L /usr/lib/systemd/network -name '*.link*' -execdir cp -r -pt "$DESTDIR/usr/lib/systemd/network/" '{}' +
> ...
> ```

I'm very sorry for spending so much time on this issue. I'll do my best to complete it within two weeks. If you have any ideas for solving it, feel free to discuss them with me — PRs are also welcome. :)

winstonma:
I'm posting this because whenever an error occurs on their system, they seem to be aware of the root cause, at least in my experience. The error typically happens when the udev script is executed after a new kernel is installed or when [Plymouth](https://wiki.ubuntu.com/Plymouth) is being upgraded. 

When the issue arises, they appear to know that a short-term solution is to reinstall the original findutils package.
```
#METHOD #1
sudo oxidizr disable --experiments findutils

# METHOD #2
sudo apt install --reinstall findutils
```

Issue #15, opened 2017-03-15
CodeHealth: tidy up Errors


mcharsley:
We currently just return errors created from arbitrary strings. At some point we should switch to a proper find::Error enum

Issue #6, opened 2017-03-15
FeatureComplete: implement exec[dir] +


mcharsley:
We (will soon) have support for -exec and -execdir clauses that end with a ';' (i.e. run this command for every file/directory). We need to also add support for clauses that end with a '+' (i.e. batch up the files/dirs and then run the command for as many as possible at once). 

Unfortunately this isn't easy in an os-independent way because the standard library doesn't expose any way of telling when a command-line is going to be too long. I raised https://github.com/rust-lang/rust/issues/40384 but it's not getting much traction.

So we need to go for a lowest common denominator approach. Choose a hard-coded limit (I'd suggest a bit less than 8kB, to allow for any inaccuracies in the next bit), come up with an efficient way of estimating the command-line length (doing this accurately is going to involve reimplementing too much of std::process::Command) and trigger the command when the estimated total goes over the limit.

To do this we need to
 - add a MultiExecMatcher class to find::matchers::exec (implementing the finished_dir and finished methods to make the exec calls for any remaining files that haven't been executed yet)
 - tweak find::process_dir to call the finished and finished_dir methods as appropriate
 - tweak find::matchers::build_matcher_tree to stop returning an error when -exec[dir] finds a + and create a MultiExecMatcher instead.

PR #411, opened 2024-07-04
Implement `-ignore_readdir_race` and `-noignore_readdir_race`.

### LINKED ISSUE #377: Implement `-ignore_readdir_race` and `-noignore_readdir_race`
refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Directories.html

```
Option: -ignore_readdir_race
If a file disappears after its name has been read from a directory but 
before find gets around to examining the file with stat, don’t issue an error message. 
If you don’t specify this option, an error message will be issued.

Furthermore, find with the ‘-ignore_readdir_race’ option will ignore errors of 
the ‘-delete’ action in the case the file has disappeared since the parent directory was read: 
it will not output an error diagnostic, and the return code of the ‘-delete’ action will be true.

This option can be useful in system scripts (cron scripts, for example) that examine areas of 
the filesystem that change frequently (mail queues, temporary directories, and so forth), 
because this scenario is common for those sorts of directories. 
Completely silencing error messages from find is undesirable, 
so this option neatly solves the problem. 
There is no way to search one part of the filesystem with this option on and 
part of it with this option off, though. 
When this option is turned on and find discovers that one of the start-point files specified on 
the command line does not exist, no error message will be issued.

Option: -noignore_readdir_race
This option reverses the effect of the ‘-ignore_readdir_race’ option.
```

hanbings:
This PR will pass the test about race conditions in BFS test. Our implementation does not seem to cause race conditions, so only add `Config` related to `-ignore_readdir_race` and `-noignore_readdir_race` without making other logical changes.

Closes #377

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -692,6 +692,14 @@ fn build_matcher_tree(
 
                 return Ok((i, top_level_matcher.build()));
             }
+            "-ignore_readdir_race" => {
+                config.ignore_readdir_race = true;
+                None
+            }
+            "-noignore_readdir_race" => {
+                config.ignore_readdir_race = false;
+                None
+            }
             "-d" | "-depth" => {
                 // TODO add warning if it appears after actual testing criterion
                 config.depth_first = true;
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -708,6 +708,7 @@ fn build_matcher_tree(
             "-noignore_readdir_race" => {
                 config.ignore_readdir_race = false;
                 None
+            }
             "-daystart" => {
                 config.today_start = true;
                 None
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -692,11 +692,11 @@ fn build_matcher_tree(
 
                 return Ok((i, top_level_matcher.build()));
             }
-            // In our implementation, including the `-exec` parameter, 
-            // it is always run in a single thread. 
+            // In our implementation, including the `-exec` parameter,
+            // it is always run in a single thread.
             // Therefore, there is no race condition for now.
             // and we currently only add the corresponding fields in Config.
-            // 
+            //
             // Related: https://github.com/uutils/findutils/pull/411#issuecomment-2210638686
             "-ignore_readdir_race" => {
                 config.ignore_readdir_race = true;
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -692,6 +692,12 @@ fn build_matcher_tree(
 
                 return Ok((i, top_level_matcher.build()));
             }
+            // In our implementation, including the `-exec` parameter, 
+            // it is always run in a single thread. 
+            // Therefore, there is no race condition for now.
+            // and we currently only add the corresponding fields in Config.
+            // 
+            // Related: https://github.com/uutils/findutils/pull/411#issuecomment-2210638686
             "-ignore_readdir_race" => {
                 config.ignore_readdir_race = true;
                 None
--- a/src/find/mod.rs
+++ b/src/find/mod.rs
@@ -34,6 +34,7 @@ impl Default for Config {
             sorted_output: false,
             help_requested: false,
             version_requested: false,
+            // For compat: doesn't change anything
             ignore_readdir_race: false,
         }
     }
--- a/src/find/mod.rs
+++ b/src/find/mod.rs
@@ -1221,4 +1221,35 @@ mod tests {
 
         assert_eq!(rc, 0);
     }
+
+    #[test]
+    #[cfg(unix)]
+    fn test_ignore_readdir_race() {
+        use crate::find::tests::FakeDependencies;
+
+        let deps = FakeDependencies::new();
+        let rc = find_main(
+            &["find", "./test_data/simple/subdir", "-ignore_readdir_race"],
+            &deps,
+        );
+
+        assert_eq!(rc, 0);
+    }
+
+    #[test]
+    fn test_noignore_readdir_race() {
+        use crate::find::tests::FakeDependencies;
+
+        let deps = FakeDependencies::new();
+        let rc = find_main(
+            &[
+                "find",
+                "./test_data/simple/subdir",
+                "-noignore_readdir_race",
+            ],
+            &deps,
+        );
+
+        assert_eq!(rc, 0);
+    }
 }
--- a/src/find/mod.rs
+++ b/src/find/mod.rs
@@ -21,6 +21,7 @@ pub struct Config {
     sorted_output: bool,
     help_requested: bool,
     version_requested: bool,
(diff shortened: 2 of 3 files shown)


sylvestre:
could you please document the fact that this option doesn't do anything ?

hanbings:
> could you please document the fact that this option doesn't do anything ?

Sorry for the late reply.

In short, taking the test https://github.com/tavianator/bfs/blob/main/tests/gnu/ignore_readdir_race.sh as an example, during the execution of `-exec`, our main thread will always be blocked by the `-exec` executor, so there will be no race condition.
But I guess this problem may not appear until https://github.com/uutils/findutils/issues/6 is implemented or https://github.com/uutils/findutils/issues/366 is fixed.

### Wirte up
I usually start by checking the GNU tests and the BFS tests for this parameter.
The test project for `-ignore_readdir_race` is at https://github.com/tavianator/bfs/blob/main/tests/gnu/ignore_readdir_race.sh

`ignore_readdir_race.sh`
```sh
cd "$TEST"
"$XTOUCH" foo bar

# -links 1 forces a stat() call, which will fail for the second file
invoke_bfs . -mindepth 1 -ignore_readdir_race -links 1 -exec "$TESTS/remove-sibling.sh" {} \;
```

`remove-sibling.sh`
```sh
#!/usr/bin/env bash

for file in "${1%/*}"/*; do
    if [ "$file" != "$1" ]; then
        rm "$file"
        exit $?
    fi
done

exit 1
```

I then ran the script on my local machine:
```
/tmp/test_dir$ touch foo bar
# and copy remove-sibling.sh
/tmp/test_dir$ ls
foo bar remove-sibling.sh

# GNU find
/tmp/test_dir$ find . -mindepth 1 -links 1 -exec "./remove-sibling.sh" {} \;
find: ‘./bar’: No such file or directory
/tmp/test_dir$ ls
remove-sibling.sh

# uutils/find
/tmp/test_dir$ touch foo bar

/tmp/test_dir$ ls
foo bar remove-sibling.sh

/tmp/test_dir$ ... findutils/target/debug/find . -mindepth 1 -links 1 -exec "./remove-sibling.sh" {} \;
/tmp/test_dir$ ls
remove-sibling.sh
```


Then I realized that our implementation didn't seem to have this problem in the same test.
OK, so let's look at the code.

First we jump here from the main function, then turns to the parser `parse_args` which parses the arguments.
https://github.com/uutils/findutils/blob/main/src/find/mod.rs#L176
```rust
fn do_find<'a>(args: &[&str], deps: &'a dyn Dependencies<'a>) -> Result<u64, Box<dyn Error>> {
    let paths_and_matcher = parse_args(args)?;
    ...
}
```

After getting all the matchers that meet the parameters, use `process_dir` to traverse the directory.
https://github.com/uutils/findutils/blob/main/src/find/mod.rs#L189
```rust
found_count += process_dir(
    &path,
    &paths_and_matcher.config,
    deps,
    &*paths_and_matcher.matcher,
    &mut quit,
);
```

Here, `matcher.matches` is executed once for each directory entry.
So far, everything has been executed linearly, and no parallel processing has occurred.
https://github.com/uutils/findutils/blob/main/src/find/mod.rs#L159
```rust
while let Some(result) = it.next() {
    match result {
        Err(err) => {
            uucore::error::set_exit_code(1);
            writeln!(&mut stderr(), "Error: {dir}: {err}").unwrap()
        }
        Ok(entry) => {
            let mut matcher_io = matchers::MatcherIO::new(deps);
            if matcher.matches(&entry, &mut matcher_io) {
                found_count += 1;
            }
            ...
        }
    }
}
```

Finally, there is the part of `exec.rs` that executes the instructions.
When executing a command, we must wait for the command to complete before returning a result of whether the execution is completed (of course, an error message is output on the screen here.)
https://github.com/uutils/findutils/blob/main/src/find/matchers/exec.rs#L87
```rust
       ...
       match command.status() {
            Ok(status) => status.success(),
            Err(e) => {
                writeln!(&mut stderr(), "Failed to run {}: {}", self.executable, e).unwrap();
                false
            }
       }
       ...
```

Therefore, in the case of `-exec`, we always complete all work in sequence within one thread, and there will be no situation where instructions are not completed but `uutils/findutils` is still traversing.

sylvestre:
oh, sorry, i just meant "in the code. explaining why it isn't doing anything" :)

hanbings:
> oh, sorry, i just meant "in the code. explaining why it isn't doing anything" :)

Ok, I've submitted the description in my latest commit.

src/find/mod.rs
```diff
@@ -21,6 +21,7 @@ pub struct Config {
     sorted_output: bool,
     help_requested: bool,
     version_requested: bool,
+    ignore_readdir_race: bool,
```
sylvestre, line 24:
```suggestion
    // For compat: doesn't change anything
    ignore_readdir_race: bool,
```

hanbings: Thanks, committed :)

src/find/matchers/mod.rs
```diff
@@ -692,6 +692,20 @@ fn build_matcher_tree(
 
                 return Ok((i, top_level_matcher.build()));
             }
+            // In our implementation, including the `-exec` parameter,
+            // it is always run in a single thread.
+            // Therefore, there is no race condition for now.
```
tavianator, line 697: The "race condition" that this flag is talking about is *another* command simultaneously modifying the directory while `find` is reading it, something like

```console
tavianator@graphene $ touch {1..10000}
tavianator@graphene $ find -size 1 & rm ./*
[1] 65991
find: ‘./10000’: No such file or directory
[1]  + 65991 exit 1     find -size 1
tavianator@graphene $ touch {1..10000}
tavianator@graphene $ find -ignore_readdir_race -size 1 & rm ./*
[1] 66034
[1]  + 66034 done       find -ignore_readdir_race -size 1
```

hanbings: Sorry, I forgot about this. The race condition does happen when deleting a large number of files, and I'm rewriting a piece of code that implements this feature.

hanbings: Terribly sorry, I did misunderstand the meaning of `readdir_race`. 

I rechecked the code and read the GNU documentation. The `-ignore_readdir_race` parameter requires a way to suppress file not found errors globally in the software, so I need to complete https://github.com/uutils/findutils/issues/15 to centralize the error handling logic before I can return here. 

I will finish them as soon as possible. :)

sylvestre:
needs rebasing

sylvestre:
@hanbings do you have an update on this ? thanks :)

hanbings:
> @hanbings do you have an update on this ? thanks :)

There is nothing new here yet, and completing this PR requires completing the global error handling associated with this feature (which is https://github.com/uutils/findutils/issues/15). 
Before that I would like to complete https://github.com/uutils/findutils/pull/369 first. :)

sylvestre:
@hanbings any plan to finish it ? thanks :)

PR #413, opened 2024-07-06
Implement `-daystart`

### LINKED ISSUE #372: Implement `-daystart`
refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Age-Ranges.html

```
Option: -daystart
Measure times from the beginning of today rather than from 24 hours ago. 
So, to list the regular files in your home directory that were modified yesterday, do

find ~/ -daystart -type f -mtime 1

The ‘-daystart’ option is unlike most other options in that it has an effect 
on the way that other tests are performed. 
The affected tests are ‘-amin’, ‘-cmin’, ‘-mmin’, ‘-atime’, ‘-ctime’ and ‘-mtime’. 
The ‘-daystart’ option only affects the behaviour of any tests which 
appear after it on the command line.
```

hanbings:
Closes #372

### DIFF
--- a/src/find/matchers/time.rs
+++ b/src/find/matchers/time.rs
@@ -7,7 +7,7 @@
 use std::error::Error;
 use std::fs::{self, Metadata};
 use std::io::{stderr, Write};
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use walkdir::DirEntry;
 
 use super::{ComparableValue, Matcher, MatcherIO};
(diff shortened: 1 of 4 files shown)


src/find/matchers/time.rs
```diff
@@ -250,44 +260,68 @@ impl Matcher for FileTimeMatcher {
...
 pub struct FileAgeRangeMatcher {
     minutes: ComparableValue,
     file_time_type: FileTimeType,
+    today_start: bool,
 }
 
 impl Matcher for FileAgeRangeMatcher {
     fn matches(&self, file_info: &DirEntry, matcher_io: &mut MatcherIO) -> bool {
-        match self.matches_impl(file_info, matcher_io.now()) {
+        let time = if self.today_start {
```
sylvestre, line 315: maybe move this into a function as you are duplicating it?

hanbings: Thanks, it would be better to wrap it in a function. Changed in the latest commit.

PR #414, opened 2024-07-08
Implement `-noleaf`

### LINKED ISSUE #376: Implement `-noleaf`
refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Directories.html

```
Option: -noleaf

Do not optimize by assuming that directories contain 2 fewer subdirectories than their hard link count. 
This option is needed when searching filesystems that do not follow the Unix directory-link convention, 
such as CD-ROM or MS-DOS filesystems or AFS volume mount points. 
Each directory on a normal Unix filesystem has at least 2 hard links: its name and its . entry. 
Additionally, its subdirectories (if any) each have a .. entry linked to that directory. 
When find is examining a directory, 
after it has statted 2 fewer subdirectories than the directory’s link count, 
it knows that the rest of the entries in the directory are non-directories (leaf files in the directory tree). 
If only the files’ names need to be examined, 
there is no need to stat them; this gives a significant increase in search speed.
```

hanbings:
We use the walkdir library for directory traversal in our codebase, and the current predicates are all secondary filtering on the results returned by walkdir. So I guess `-noleaf` will not bring any optimization (unless `walkdir` includes this optimization, to be honest I don't know much about `walkdir` and didn't find any related functions in their documentation).

The current code only contains compatibility code.

Closes #376

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -692,6 +692,10 @@ fn build_matcher_tree(
 
                 return Ok((i, top_level_matcher.build()));
             }
+            "-noleaf" => {
+                config.no_leaf_dirs = true;
+                None
+            }
             "-d" | "-depth" => {
                 // TODO add warning if it appears after actual testing criterion
                 config.depth_first = true;
--- a/src/find/mod.rs
+++ b/src/find/mod.rs
@@ -21,6 +21,7 @@ pub struct Config {
     sorted_output: bool,
     help_requested: bool,
     version_requested: bool,
+    no_leaf_dirs: bool,
 }
 
 impl Default for Config {
@@ -33,6 +34,7 @@ impl Default for Config {
             sorted_output: false,
             help_requested: false,
             version_requested: false,
+            no_leaf_dirs: false,
         }
     }
 }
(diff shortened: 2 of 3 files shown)


src/find/matchers/mod.rs
```diff
@@ -692,6 +692,10 @@ fn build_matcher_tree(
 
                 return Ok((i, top_level_matcher.build()));
             }
+            "-noleaf" => {
+                config.no_leaf_dirs = true;
```
sylvestre, line 696:
```suggestion
                // No change of behavior
                config.no_leaf_dirs = true;
```

hanbings: Thanks, committed :)

sylvestre:
LGTM
Maybe we should create a doc like compatibility.md to keep track of these flags that aren't doing anything but here for compat. WDYT ?

hanbings:
> LGTM Maybe we should create a doc like compatibility.md to keep track of these flags that aren't doing anything but here for compat. WDYT ?

Sound like a perfect idea, we cannot assume that every user will review the code to determine compatibility.
I'll submit them in a new PR later. :)

PR #416, opened 2024-07-10
find: ctime is "changed" time, not "created" time

### LINKED ISSUE #358: find: The behavior of the parameter `-ctime` is inconsistent with the GNU manual description.
I found that the current interpretation of `-ctime` in the code is `Created` time, but the interpretation of `-c` in the GNU manual seems to be `Status Changed`. 

The following is a piece of relevant code: https://github.com/uutils/findutils/blob/main/src/find/matchers/mod.rs#L412
Manual for the -ctime: https://www.gnu.org/software/findutils/manual/html_node/find_html/Age-Ranges.html#Age-Ranges

tavianator:
Fixes #358.


hanbings (approved):

hanbings:
Merged, thanks!

PR #417, opened 2024-07-10
options return true

tavianator:
- **find: Options like -maxdepth return true**
- **find: Stop parsing after we see -help or -version**


hanbings (approved):
LGTM! thanks.

PR #420, opened 2024-07-16
Implement `-follow`

### LINKED ISSUE #308: Implement `-follow`
Even if this is deprecated, for compat, we should implement it:
```

       -follow
              Deprecated; use the  -L  option  instead.   Dereference  symbolic  links.   Implies
              -noleaf.   The -follow option affects only those tests which appear after it on the
              command line.  Unless the -H or -L option has been specified, the position  of  the
              -follow  option  changes the behaviour of the -newer predicate; any files listed as
              the argument of -newer will be dereferenced if they are symbolic links.   The  same
              consideration applies to -newerXY, -anewer and -cnewer.  Similarly, the -type pred-
              icate will always match against the type of the file that a symbolic link points to
              rather  than  the  link itself.  Using -follow causes the -lname and -ilname predi-
              cates always to return false.
```

hanbings:
Closed #308

May conflict with https://github.com/uutils/findutils/pull/416, better merge #416 first. :)

There are changes as described in `man find`:
(The following changes are: When matching files, match the file pointed to by the file link.)

1. It will use noleaf by default. (But -noleaf has no behavior change yet)
2. changes the behaviour of the -newer predicate.
3. consideration applies to -newerXY, -anewer and -cnewer
4. -type predicate will always match against the type of the file that a symbolic link points to rather than the link itself.
5. causes the -lname and -ilname predicates always to return false.

### DIFF
--- a/src/find/matchers/lname.rs
+++ b/src/find/matchers/lname.rs
@@ -37,23 +37,17 @@ fn read_link_target(file_info: &DirEntry) -> Option<PathBuf> {
 /// pattern. See `glob::Pattern` for details on the exact syntax.
 pub struct LinkNameMatcher {
     pattern: Pattern,
-    follow: bool,
 }
 
 impl LinkNameMatcher {
-    pub fn new(pattern_string: &str, caseless: bool, follow: bool) -> LinkNameMatcher {
+    pub fn new(pattern_string: &str, caseless: bool) -> LinkNameMatcher {
         let pattern = Pattern::new(pattern_string, caseless);
-        Self { pattern, follow }
+        Self { pattern }
     }
 }
 
 impl Matcher for LinkNameMatcher {
     fn matches(&self, file_info: &DirEntry, _: &mut MatcherIO) -> bool {
-        // -follow: causes the -lname and -ilname predicates always to return false.
-        if self.follow {
-            return false;
-        }
-
         if let Some(target) = read_link_target(file_info) {
             self.pattern.matches(&target.to_string_lossy())
         } else {
@@ -98,7 +92,7 @@ mod tests {
         create_file_link();
 
         let link_f = get_dir_entry_for("test_data/links", "link-f");
-        let matcher = LinkNameMatcher::new("ab?bc", false, false);
+        let matcher = LinkNameMatcher::new("ab?bc", false);
         let deps = FakeDependencies::new();
         assert!(matcher.matches(&link_f, &mut deps.new_matcher_io()));
     }
@@ -108,19 +102,8 @@ mod tests {
         create_file_link();
 
         let link_f = get_dir_entry_for("test_data/links", "link-f");
-        let matcher = LinkNameMatcher::new("AbB?c", true, false);
+        let matcher = LinkNameMatcher::new("AbB?c", true);
         let deps = FakeDependencies::new();
         assert!(matcher.matches(&link_f, &mut deps.new_matcher_io()));
-    }
-
-    #[test]
-    fn matches_with_follow_option() {
-        create_file_link();
-
-        let link_f = get_dir_entry_for("test_data/links", "link-f");
-        let matcher = LinkNameMatcher::new("ab?bc", false, true);
-        let deps = FakeDependencies::new();
-        // This matcher always returns false when the follow option is enabled.
-        assert!(!matcher.matches(&link_f, &mut deps.new_matcher_io()));
     }
 }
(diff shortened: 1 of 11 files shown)


hanbings:
Sorry, the test coverage is still not up to target. 
They almost always appear in error handling logic and, therefore, are difficult to cover by test code.
I would like to focus on improving coverage after completing all GNU feature implementations.

src/find/matchers/lname.rs
```diff
@@ -37,17 +37,23 @@ fn read_link_target(file_info: &DirEntry) -> Option<PathBuf> {
...
-        Self { pattern }
+        Self { pattern, follow }
     }
 }
 
 impl Matcher for LinkNameMatcher {
     fn matches(&self, file_info: &DirEntry, _: &mut MatcherIO) -> bool {
+        // -follow: causes the -lname and -ilname predicates always to return false.
+        if self.follow {
+            return false;
```
tavianator, line 54: Not really:

```console
tavianator@graphene $ ln -s nowhere broken
tavianator@graphene $ find -lname nowhere
./broken
```

hanbings: Sorry, could you please give me some more detailed steps?
I noticed that the `-follow` parameter is missing, is that `find -follow -lname nowhere`?

tavianator: Whoops I copy pasted the wrong part of my session, but yeah:

```console
$ ln -s nowhere broken
$ find -follow -lname nowhere
./broken
```

hanbings: Thanks. I have fixed it to the actual behavior of the GNU version in https://github.com/uutils/findutils/pull/420/commits/a0d4e468676299fde7d8ae63108864a71f9770b4

tavianator (changes requested):

src/find/matchers/time.rs
```diff
@@ -29,23 +29,54 @@ fn get_time(matcher_io: &mut MatcherIO, today_start: bool) -> SystemTime {
...
 /// This matcher checks whether a file is newer than the file the matcher is initialized with.
 pub struct NewerMatcher {
     given_modification_time: SystemTime,
+    follow: bool,
 }
 
 impl NewerMatcher {
-    pub fn new(path_to_file: &str) -> Result<Self, Box<dyn Error>> {
+    pub fn new(path_to_file: &str, follow: bool) -> Result<Self, Box<dyn Error>> {
         let metadata = fs::metadata(path_to_file)?;
```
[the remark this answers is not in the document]
hanbings: Thanks, I opened an issue tracking this: https://github.com/uutils/findutils/issues/431, and I'd like to open some new PRs to address it. :)

src/find/matchers/time.rs
```diff
@@ -29,23 +29,54 @@ fn get_time(matcher_io: &mut MatcherIO, today_start: bool) -> SystemTime {
...
+            follow,
         })
     }
 
     /// Implementation of matches that returns a result, allowing use to use try!
     /// to deal with the errors.
     fn matches_impl(&self, file_info: &DirEntry) -> Result<bool, Box<dyn Error>> {
-        let this_time = file_info.metadata()?.modified()?;
+        let path = file_info.path();
+        let this_time = if self.follow && path.is_symlink() {
```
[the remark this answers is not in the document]
hanbings: The corresponding logic for using `symlink_metadata` has been added in the current commit, and I have also opened an issue to track the problem of `metadata` caching: https://github.com/uutils/findutils/issues/430. Thanks!

src/find/matchers/lname.rs
```diff
@@ -27,48 +33,43 @@ fn read_link_target(file_info: &DirEntry) -> Option<PathBuf> {
                 )
                 .unwrap();
             }
-
```
tavianator, line 30: Should probably keep all these deleted blank lines

tavianator (changes requested):
Presumably you want to call [`WalkDir::follow_links(config.follow)`](https://docs.rs/walkdir/latest/walkdir/struct.WalkDir.html#method.follow_links) at some point?

Also, you could implement `-L` at the same time.

tavianator (approved):

tavianator:
@sylvestre #436 contains a squashed version of this PR if you'd rather avoid fixing the tests here

tavianator:
This can be closed now that #436 was merged

PR #421, opened 2024-07-20
Implement `-fprint`

### LINKED ISSUE #381: Implement `-fprint`
refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Print-File-Name.html

```
Action: -fprint file

True; print the entire file name into file file, followed by a newline. 
If file does not exist when find is run, it is created; if it does exist, it is truncated to 0 bytes. 
The named output file is always created, even if no output is sent to it. 
The file names /dev/stdout and /dev/stderr are handled specially; 
they refer to the standard output and standard error output, respectively.

If there is the faintest possibility that 
one of the files for which you are searching might contain a newline, 
you should use ‘-fprint0’ instead.
```

hanbings:
Closed #381


src/find/matchers/mod.rs
```diff
@@ -339,6 +340,23 @@ fn parse_str_to_newer_args(input: &str) -> Option<(String, String)> {
...
+/// Creates a file if it doesn't exist.
+/// If it does exist, it will be overwritten.
+fn get_or_create_file(path: &str) -> Result<File, Box<dyn Error>> {
+    let file = match File::open(path) {
+        Ok(file) => file,
+        Err(err) => {
+            if err.kind() == std::io::ErrorKind::NotFound {
+                File::create(path)?
+            } else {
+                return Err(From::from(err));
```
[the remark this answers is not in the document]
hanbings: I checked the [ErrorKind](https://doc.rust-lang.org/std/io/enum.ErrorKind.html) page, how about we select PermissionDenied as a test case and use the following code to trigger it under Unix system?

```
#[cfg(unix)]
{
      let result = get_or_create_file("/etc/shadow");
      assert!(result.is_err());
}
```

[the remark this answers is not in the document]
hanbings: I've added it at https://github.com/uutils/findutils/pull/421/commits/823949cceb729b4b3bdb15ad13d290846d9256aa I think this is the most stable trigger we can get besides nightly features.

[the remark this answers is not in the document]
tavianator: Why don't you just use `File::create()`?  That truncates the file, but so does `-fprint`:

```console
tavianator@tachyon $ echo Hello world >foo
tavianator@tachyon $ find -fprint foo -prune
tavianator@tachyon $ cat foo
.
```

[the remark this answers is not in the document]
hanbings: Oops... I didn't notice that create includes overwriting existing files. Corrected the code in the new commit. Thanks.

src/find/matchers/mod.rs
```diff
@@ -339,6 +340,19 @@ fn parse_str_to_newer_args(input: &str) -> Option<(String, String)> {
     }
 }
 
+/// Creates a file if it doesn't exist.
+/// If it does exist, it will be overwritten.
+fn get_or_create_file(path: &str) -> Result<File, Box<dyn Error>> {
```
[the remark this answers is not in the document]
hanbings: Would it be better to give the user a more friendly prompt here? But yes, it is better to use "?" for now.
I would like to keep `get_or_create_file` function first, so that it is convenient to modify these unexpected behaviors in a unified way later. :)
Changes have been committed at https://github.com/uutils/findutils/pull/421/commits/d02b2c491f87f00610c62b662c2b13a2ac83b319. thanks.

tests/find_cmd_tests.rs
```diff
@@ -946,3 +946,25 @@ fn find_daystart() {
         .success()
         .stderr(predicate::str::is_empty());
 }
+
+#[test]
+#[serial(working_dir)]
+fn find_nogroup() {
```
tavianator, line 952: Why is this called `nogroup`?  There is a `-nogroup` primary, but this isn't testing it.

src/find/matchers/printer.rs
```diff
@@ -25,25 +27,44 @@ impl std::fmt::Display for PrintDelimiter {
...
-        let mut out = matcher_io.deps.get_output().borrow_mut();
-        write!(
-            out,
-            "{}{}",
-            file_info.path().to_string_lossy(),
-            self.delimiter
-        )
-        .unwrap();
-        out.flush().unwrap();
+        match &self.output_file {
```
tavianator, line 44: You can merge the two similar blocks with something like

```rust
    fn print(&self, file_info: &DirEntry, mut out: impl Write) {
        write!(out, "{}{}", file_info.path().to_string_lossy(), self.delimiter).unwrap();
        out.flush().unwrap();
    }

    fn matches(&self, file_info: &DirEntry, matcher_io: &mut MatcherIO) -> bool {
        if let Some(file) = &self.output_file {
            self.print(file_info, file);
        } else {
            self.print(file_info, &mut *matcher_io.deps.get_output().borrow_mut());
        }
        true
    }
```

src/find/matchers/printer.rs
```diff
@@ -25,25 +27,44 @@ impl std::fmt::Display for PrintDelimiter {
...
+            Some(output_file) => {
+                let mut out = output_file;
+                write!(
+                    out,
+                    "{}{}",
+                    file_info.path().to_string_lossy(),
+                    self.delimiter
+                )
+                .unwrap();
```
tavianator, line 53: GNU find doesn't treat this as a fatal error.  You can try `-fprint /dev/full` on Linux for example.

src/find/matchers/printer.rs
```diff
@@ -25,25 +27,44 @@ impl std::fmt::Display for PrintDelimiter {
...
+                out.flush().unwrap();
```
tavianator, line 54: Flushing after every path is probably too slow.  GNU find just uses the stream's default buffering, i.e. line-buffered for TTYs and fully buffered for everything else.

tavianator (changes requested):

src/find/matchers/printer.rs
hanbings: Thanks, the other comment in the current review have been revised.
and GNU find seems to also interrupt the traversal of the remaining files. Should we also interrupt the traversal and give a suitable message?

hanbings: Thanks, changed in new commit.

src/find/matchers/printer.rs
```diff
@@ -25,25 +30,53 @@ impl std::fmt::Display for PrintDelimiter {
...
+            Err(e) => {
+                if print_error_message {
+                    writeln!(
+                        &mut stderr(),
+                        "Error writing {:?} for {}",
+                        file_info.path().to_string_lossy(),
+                        e
+                    )
+                    .unwrap();
+                    uucore::error::set_exit_code(1);
```
tavianator, line 61: This could be a method on `MatcherIO`, and then you wouldn't need to unset it in the tests

src/find/matchers/mod.rs
```diff
@@ -361,15 +369,24 @@ fn build_matcher_tree(
...
                 i += 1;
                 Some(Printf::new(args[i])?.into_box())
             }
+            "-fprint" => {
+                if i >= args.len() - 1 {
+                    return Err(From::from(format!("missing argument to {}", args[i])));
+                }
+                i += 1;
+
+                let file = get_or_create_file(args[i])?;
```
tavianator, line 387: For future reference: GNU find de-duplicates files, so if you do `find -fprint foo -fprint foo` it will use the same `FILE *` for both and you'll get each line twice.  Right now uutils will use two different `File`s and the writes will overlap with each other.

tavianator (approved):

sylvestre:
@hanbings could you please address the two last comments in new PR ?

PR #435, opened 2024-08-07
Implement `-ls` and `-fls`

### LINKED ISSUE #382: Implement `-ls` and `-fls`
refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Print-File-Information.html
**Note that this option has many different cases, if we are not familiar with this we will need to consult the reference.**

```
Action: -ls
True; list the current file in ‘ls -dils’ format on the standard output. The output looks like this:

204744   17 -rw-r--r--   1 djm      staff       17337 Nov  2  1992 ./lwall-quotes

Action: -fls file
True; like ‘-ls’ but write to file like ‘-fprint’. 
The named output file is always created, even if no output is sent to it.
```

`-fprint` see [Print File Name](https://www.gnu.org/software/findutils/manual/html_node/find_html/Print-File-Name.html)

hanbings:
Closed #382

### DIFF
--- a/src/find/matchers/ls.rs
+++ b/src/find/matchers/ls.rs
@@ -8,9 +8,8 @@ use std::{
     fs::File,
     io::{stderr, Write},
 };
-use walkdir::DirEntry;
 
-use super::{Matcher, MatcherIO};
+use super::{Matcher, MatcherIO, WalkEntry};
 
 #[cfg(unix)]
 fn format_permissions(mode: uucore::libc::mode_t) -> String {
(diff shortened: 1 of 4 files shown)


src/find/matchers/ls.rs
```diff
@@ -0,0 +1,256 @@
...
+    fs::File,
+    io::{stderr, Write},
+};
+use walkdir::DirEntry;
+
+use super::{Matcher, MatcherIO};
+
+#[cfg(unix)]
+fn format_permissions(mode: u32) -> String {
+    let file_type = if mode & 0o170000 == 0o040000 {
```
tavianator, line 17: It would be better to use symbolic constants like `mode & libc::S_IFMT == libc::S_IFDIR`

hanbings: thanks. fixed them in the new commit.

sylvestre:
needs rebasing

PR #436, opened 2024-08-08
Wrap walkdir::DirEntry in a new type

### LINKED ISSUE #12: FeatureComplete: handle symlinks
This involves handling -H, -L and finishing find::matchers::type_matcher.

Take a look at https://github.com/BurntSushi/walkdir/blob/master/examples/walkdir.rs for hints on how to handle -L with walkdir and https://github.com/BurntSushi/walkdir/blob/master/src/tests.rs on how to test interaction with symlinks on both unix and windows machines

### LINKED ISSUE #308: Implement `-follow`

### LINKED ISSUE #379: Implement `-xtype`
refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Type.html

```
Test: -xtype c

This test behaves the same as ‘-type’ unless the file is a symbolic link. If the file is a symbolic link, 
the result is as follows (in the table below, ‘X’ should be understood to represent any letter except ‘l’):

‘‘-P -xtype l’’
True if the symbolic link is broken

‘‘-P -xtype X’’
True if the (ultimate) target file is of type ‘X’.

‘‘-L -xtype l’’
Always true

‘‘-L -xtype X’’
False unless the symbolic link is broken
```

### LINKED ISSUE #412: Implement `-P` `-H` and `-L`
refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Symbolic-Links.html

```
‘-P’
find does not dereference symbolic links at all. This is the default behaviour. 
This option must be specified before any of the file names on the command line.

‘-H’
find does not dereference symbolic links (except in the case of file names on the command line, 
which are dereferenced). If a symbolic link cannot be dereferenced, 
the information for the symbolic link itself is used. 
This option must be specified before any of the file names on the command line.

‘-L’
find dereferences symbolic links where possible, 
and where this is not possible it uses the properties of the symbolic link itself. 
This option must be specified before any of the file names on the command line. 
Use of this option also implies the same behaviour as the ‘-noleaf’ option. 
If you later use the ‘-H’ or ‘-P’ options, this does not turn off ‘-noleaf’.

Actions that can cause symbolic links to become broken while ‘find’ is 
executing (for example ‘-delete’) can give rise to confusing behaviour. 
Take for example the command line ‘find -L . -type d -delete’. This will delete empty directories. 
If a subtree includes only directories and symbolic links to directoires, 
this command may still not successfully delete it, 
since deletion of the target of the symbolic link will 
cause the symbolic link to become broken and ‘-type d’ is false for broken symbolic links.
```

### LINKED ISSUE #440: Centralize handling of `uucore::error::set_exit_code()` calls.
refer: https://github.com/uutils/findutils/pull/421#discussion_r1705629746

> This could be a method on `MatcherIO`, and then you wouldn't need to unset it in the tests

Note: Currently we have two `uucore::error::set_exit_code()` in the entire find code.

tavianator:
Includes #420.

Closes #12.  Closes #308.  Closes #379.  Closes #412.  Closes #440.  Implements the `Metadata` caching part of #430.


src/find/matchers/mod.rs
```diff
@@ -712,9 +795,24 @@ fn build_matcher_tree(
...
+                // 5. causes the -lname and -ilname predicates always to return false.
+                //    (unless they happen to match broken symbolic links)
+                config.follow = Follow::Always;
+                config.no_leaf_dirs = true;
+                Some(TrueMatcher.into_box())
+            }
             "-daystart" => {
                 config.today_start = true;
-                Some(TrueMatcher.into_box())
+                None
```
hanbings, line 815: I remember that this was changed to `Some(TrueMatcher.into_box())` not long ago, why was it changed back?

src/find/matchers/type_matcher.rs
```diff
@@ -5,70 +5,89 @@
...
+        "l" => FileType::Symlink,
+        "b" => FileType::BlockDevice,
+        "c" => FileType::CharDevice,
+        "p" => FileType::Fifo, // named pipe (FIFO)
+        "s" => FileType::Socket,
+        // D: door (Solaris)
+        "D" => {
+            return Err(From::from(format!(
+                "Type argument {type_string} not supported yet"
+            )))
```
hanbings, line 29: Should we open a new issue to track this? [The GNU documentation](https://www.gnu.org/software/findutils/manual/html_node/find_html/Type.html) only describes `-type D` as `door (Solaris)`.

src/find/matchers/mod.rs
tavianator: Whoops, must have been a mistake in rebasing, good catch!

src/find/matchers/type_matcher.rs
tavianator: Doors are an old IPC mechanism on Solaris.  They get their own file type.

tavianator: I've never used them but there are some old articles describing them, e.g. http://www.kohala.com/start/papers.others/doors.html

I'm not sure whether they're still in use on modern Solaris/Illumos.  I also don't think Rust's stdlib has any support for detecting doors.  I personally wouldn't worry about them.  This code was just moved from here: https://github.com/uutils/findutils/blob/465856ce2454605f7acacb0c8d4189bdbfc1c3d6/src/find/matchers/type_matcher.rs#L33-L37

hanbings (approved):

src/find/matchers/type_matcher.rs
hanbings: thanks! It is indeed old and difficult to achieve, let it stay that way. :)

hanbings:
Will there be more updates here? Should we merge it? 
Anyway this is a legendary PR, thanks! :)

tavianator:
This should be good to go!  Thanks

PR #444, opened 2024-08-12
Implement `-fprintf`

### LINKED ISSUE #383: Implement `-fprintf`
refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Print-File-Information.html

```
Action: -fprintf file format
True; like ‘-printf’ but write to file like ‘-fprint’. 
The output file is always created, even if no output is ever sent to it.
```

`-fprint` see [Print File Name](https://www.gnu.org/software/findutils/manual/html_node/find_html/Print-File-Name.html)

hanbings:
Closes #383

### DIFF
--- a/src/find/matchers/printf.rs
+++ b/src/find/matchers/printf.rs
@@ -4,11 +4,11 @@
 // license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
-use std::borrow::Cow;
 use std::error::Error;
-use std::fs;
+use std::fs::{self, File};
 use std::path::Path;
 use std::time::SystemTime;
+use std::{borrow::Cow, io::Write};
 
 use chrono::{format::StrftimeItems, DateTime, Local};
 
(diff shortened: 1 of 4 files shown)


sylvestre:
needs rebasing

sylvestre:
Any idea why `PASS +0` in ?
`Run GNU findutils tests: Changes from main: PASS +0 / FAIL -4 / ERROR +0 / SKIP +0 `

hanbings:
> Any idea why `PASS +0` in ? `Run GNU findutils tests: Changes from main: PASS +0 / FAIL -4 / ERROR +0 / SKIP +0 `

I guess it's something to do with the total number of GNU tests changing?
I haven't seen anything like this either, it seems to have been like this since the beginning of this PR.

src/find/matchers/printf.rs
```diff
@@ -619,6 +617,16 @@ impl Matcher for Printf {
...
+    }
+}
+
+impl Matcher for Printf {
+    fn matches(&self, file_info: &WalkEntry, matcher_io: &mut MatcherIO) -> bool {
+        if let Some(file) = &self.output_file {
+            self.print(file_info, file);
+        } else {
+            self.print(file_info, &mut *matcher_io.deps.get_output().borrow_mut());
+        }
```
sylvestre, line 629: a bit more rust idiomatic:

```suggestion
      let output = self.output_file.as_ref()
        .unwrap_or_else(|| matcher_io.deps.get_output().borrow_mut());
    self.print(file_info, output);
```

hanbings: Wow, that's much cleaner.

I tried modifying it but I seem to be having some type issues.

```
error[E0308]: mismatched types
   --> src/find/matchers/printf.rs:628:32
    |
628 |             .unwrap_or_else(|| matcher_io.deps.get_output().borrow_mut());
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
                                     expected `&File`, found `RefMut<'_, dyn Write>`
    |
    = note: expected reference `&File`
                  found struct `RefMut<'_, (dyn std::io::Write + 'static)>`
```

and I also tried:
```
fn print(&self, file_info: &WalkEntry, out: &mut dyn Write) { ... }

...
let output: &mut dyn Write = self.output_file
    .as_ref()
    .map(|file| file.borrow_mut() as &mut dyn Write)
    .unwrap_or_else(|| matcher_io.deps.get_output().borrow_mut());

self.print(file_info, output);
...
```

 I'm relatively new at this, would you mind pointing me out further?

sylvestre:
this version is good enough, thanks :)

PR #450, opened 2024-09-06
find: use array instead of closure

cakebaker:
This PR fixes a warning from the [manual_pattern_char_comparison](https://rust-lang.github.io/rust-clippy/master/index.html#/manual_pattern_char_comparison) lint introduced with Rust 1.81.


sylvestre:
much better!

PR #455, opened 2024-09-22
Print help/version without error prefix

andrewliebenow:


sylvestre:
please improve the commit explaining what you are fixing

andrewliebenow:
I updated the commit. New commit message:

```
xargs: print help/version without error prefix

xargs, when executed with -h/--help or -V/--version, is printing an
"Error: " prefix before the help and version text. The exit code is
also incorrect (1, instead of 0).

Display the help/version text without the erroneous prefix, and exit
with the correct exit code (success, 0).

Add tests for this bug to prevent regressions.
```

tests/xargs_tests.rs
```diff
@@ -564,3 +564,37 @@ fn xargs_replace_multiple_lines() {
...
+
+#[test]
+fn xargs_help() {
+    for option_style in ["-h", "--help"] {
+        Command::cargo_bin("xargs")
+            .expect("found binary")
+            .args([option_style])
+            .assert()
+            .success()
+            .code(0)
```
sylvestre, line 576: don't bother updating it but I think success implies code(0)

PR #457, opened 2024-09-25
find/glob: Don't use `$.` as a non-matching regex

tavianator:
The oniguruma version built by onig_sys does not include the commit
6f7b0aa ("add ONIG_SYN_BRE_ANCHOR_AT_EDGE_OF_SUBEXP into
ONIG_SYNTAX_POSIX_BASIC and ONIG_SYNTAX_GREP"), causing `$.` to actually
match a literal newline.

Using `$.` is not portable anyway; POSIX says

> A &lt;dollar-sign> ( '$' ) shall be an anchor when used as the last
> character of an entire BRE.


hanbings (approved):
thanks!

PR #470, opened 2024-11-08
ci: use `-Cinstrument-coverage` instead of `-Zprofile`

cakebaker:
This PR is an attempt to fix the "code coverage" job by using `-Cinstrument-coverage` instead of `-Zprofile` which has been removed from Rust nightly: https://github.com/rust-lang/rust/pull/131829 . For this purpose, [`cargo-llvm-cov`](https://github.com/taiki-e/cargo-llvm-cov) is used instead of `grcov`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,49 +105,23 @@ jobs:
 
       - name: Install `rust` toolchain
         run: |
-          ## Install `rust` toolchain
           rustup toolchain install nightly --no-self-update -c rustfmt --profile minimal
           rustup default nightly
 
-      - name: "`grcov` ~ install"
-        run: cargo install grcov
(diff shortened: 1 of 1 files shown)


sylvestre:
grcov also supports -Cinstrument-coverage
what was wrong with this?

they use it upstream
https://github.com/mozilla/grcov/blob/9cc638baf2cca4368b9f8354e78c1ae28dc2d168/.github/workflows/CICD.yml#L84-L98

cakebaker:
Yes, I know that `grcov` supports `-Cinstrument-coverage`, I used it in the other projects ;-)  In this project it interfered with the tests, though. I will try again with the example you provided.

sylvestre:
I know upstream ( @marco-c  ) very well ;) 
I am sure he will be happy to help

PR #498, opened 2025-02-28
find: fix "unused import" warning on Windows

cakebaker:
This PR fixes an "unused import" warning on Windows, see, for example, https://github.com/uutils/findutils/actions/runs/13512703858/job/37755835852?pr=496#step:6:181.


hanbings:
Thanks!

Issue #24, opened 2017-06-14
-name / should match ///


tavianator:
Repeated slashes for the root directory should match a single slash:

```
$ ./target/debug/find /// -maxdepth 0 -name /
$ find /// -maxdepth 0 -name /
find: warning: Unix filenames usually don't contain slashes (though pathnames do).  That means that '-name ‘/’' will probably evaluate to false all the time on this system.  You might find the '-wholename' test more useful, or perhaps '-samefile'.  Alternatively, if you are using GNU grep, you could use 'find ... -print0 | grep -FzZ ‘/’'.
///
```

hanbings:
Reopen: https://github.com/uutils/findutils/pull/508#discussion_r2012669007

PR #508, opened 2025-03-20
-name / should match ///

Every2:
Hi, I did some tests and the problem in https://github.com/uutils/findutils/issues/24 is because when we pass the './target/debug/find /// -maxdepth 0 -name /' it returns false. 
```rust
impl Matcher for NameMatcher {
         let name = file_info.file_name().to_string_lossy();
        self.pattern.matches(&name)
    }
}
```
I put an early return as an alternative solution, but I don't know if it's a good solution.

### DIFF
--- a/src/find/matchers/name.rs
+++ b/src/find/matchers/name.rs
@@ -23,6 +23,9 @@ impl NameMatcher {
 impl Matcher for NameMatcher {
     fn matches(&self, file_info: &WalkEntry, _: &mut MatcherIO) -> bool {
         let name = file_info.file_name().to_string_lossy();
+        if name.contains('/') {
+            return true;
+        }
         self.pattern.matches(&name)
     }
 }


sylvestre:
Could you please add a test to make sure we don't regress? Thanks

Every2:
> Could you please add a test to make sure we don't regress? Thanks

Of course!

Every2:
@sylvestre Added the new test, can you review it?

hanbings:
Hello, based on the Action logs, this RP seems to have caused a Windows test failure. Could you take a look? 
Added: If this functionality doesn't work on Windows, you could add a suitable macro along with a comment explaining it.
Thanks!

sylvestre:
please also add a test in tests/find_cmd_tests.rs

Every2:
@sylvestre and @hanbings Could you review the tests? Thanks for the help and guidance of both. :)

hanbings:
Thanks!

src/find/matchers/name.rs
```diff
@@ -23,6 +23,12 @@ impl NameMatcher {
 impl Matcher for NameMatcher {
     fn matches(&self, file_info: &WalkEntry, _: &mut MatcherIO) -> bool {
         let name = file_info.file_name().to_string_lossy();
+
+        #[cfg(unix)]
+        if name.contains('/') {
+            return true;
```
tavianator, line 29: This is not the right behaviour:

```console
$ target/debug/find / -maxdepth 0 -name asdf
/
```

The right behaviour is:

```console
$ find / -maxdepth 0 -name /
/
$ find /// -maxdepth 0 -name /
///
$ find / -maxdepth 0 -name ///
$ find /// -maxdepth 0 -name ///
$
```

Every2: Hi, thanks for the feedback. I'll work on it.

Every2: I would love it if you could check out the new PR.

Issue #378, opened 2024-05-11
Implement `-files0-from`


hanbings:
refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Starting-points.html

```
Option: -files0-from file

Read the starting points from file instead of getting them on the command line. 
In contrast to the known limitations of passing starting points via arguments on the command line, 
namely the limitation of the amount of file names, 
and the inherent ambiguity of file names clashing with option names, 
using this option allows to safely pass an arbitrary number of starting points to find.

Using this option and passing starting points on the command line is mutually exclusive, 
and is therefore not allowed at the same time.

The file argument is mandatory. 
One can use ‘-files0-from -’ to read the list of starting points from the standard input stream, 
and e.g. from a pipe.
```

PR #509, opened 2025-03-22
Implement : -files0-from (#378)

Crypto-Darth:
Implements : #378 

Huge Thanks to @hanbings for the implementation guide.


Crypto-Darth:
I will add more tests as soon as possible.

Crypto-Darth:
Added more tests.
I believe this much is sufficient for this function.

src/find/mod.rs
```diff
@@ -377,6 +436,36 @@ mod tests {
         }
     }
 
+    #[test]
+    fn parse_files0_basic() {
+        let parse = super::parse_args(&["-files0-from"]);
+        if let Err(e) = parse {
+            assert_eq!(e.to_string(), "missing argument to -files0-from");
+        } else {
+            panic!("parse_args should have returned an error");
```
sylvestre, line 445: please don't use panic, you should handle the error properly

src/find/mod.rs
```diff
@@ -377,6 +436,36 @@ mod tests {
...
+                    e.to_string(),
+                    "The system cannot find the file specified. (os error 2)"
+                );
+            }
+            #[cfg(unix)]
+            {
+                assert_eq!(e.to_string(), "No such file or directory (os error 2)");
+            }
+        } else {
+            panic!("parse_args should have returned an error");
```
sylvestre, line 465: same, please don't use panic

src/find/mod.rs
Crypto-Darth: Indeed but I'm a bit unsure of what to use in place of panic! , should I use unreachable! ? since the else arm should not be reached.

Crypto-Darth:
I have migrated the tests so as to avoid using panic! ,works as intended.
Please review @sylvestre .

Thank you.

src/find/mod.rs
```diff
@@ -112,24 +116,53 @@ fn parse_args(args: &[&str]) -> Result<ParsedInfo, Box<dyn Error>> {
...
-        && args[i] != "!"
-        && args[i] != "("
-    {
-        paths.push(args[i].to_string());
-        i += 1;
-    }
-    if i == paths_start {
-        paths.push(".".to_string());
+    // Check to see if '-files0-from' argument is used.
+    match config.from_file.is_some() {
```
sylvestre, line 145: why not using an if else here ? :)

Crypto-Darth: Fixed :)
not sure why I didn't use if else :) (perhaps I'm into using match more)
Thank you.

src/find/mod.rs
```diff
@@ -138,6 +168,48 @@ fn parse_args(args: &[&str]) -> Result<ParsedInfo, Box<dyn Error>> {
...
+        // empty starting point if detected shall make the program exit with non-zero code (as per GNU Manual)
+        if string_segments.iter().any(|s| s.is_empty()) {
+            return Err("Empty starting point detected in -files0-from input".into());
+        } else {
+            paths.extend(string_segments);
+        }
+    } else {
+        let file = std::fs::read(config.from_file.as_ref().unwrap())?;
+        for path in file
+            .split(|&b| b == 0)
```
tavianator, line 195: So the docs say (https://doc.rust-lang.org/std/primitive.slice.html#method.split)

> If the first element is matched, an empty slice will be the first item returned by the iterator. Similarly, if the last element in the slice is matched, an empty slice will be the last item returned by the iterator:

Meaning this is broken:

```console
$ printf 'file\0' > input
$ find -files0-from input
Error: Empty starting point detected in -files0-from input OR File is empty
```

Crypto-Darth: I have fixed it with my latest commits and seems to work as intended.
image

Thank you for the review.

src/find/mod.rs
```diff
@@ -138,6 +168,55 @@ fn parse_args(args: &[&str]) -> Result<ParsedInfo, Box<dyn Error>> {
...
+        // empty starting point if detected shall make the program exit with non-zero code (as per GNU Manual)
+        if string_segments.iter().any(|s| s.is_empty()) {
+            return Err("Empty starting point detected in -files0-from input".into());
+        } else {
+            paths.extend(string_segments);
+        }
+    } else {
+        let file = std::fs::read(config.from_file.as_ref().unwrap())?;
+        let mut file_split: Vec<&[u8]> = file.split(|&b| b == 0).collect();
+        // incase the file starts/ends with ASCII NUL
```
tavianator, line 195: I think you only need to handle ends-with:

```console
tavianator@tachyon $ printf '\0foo\0' | find -files0-from -
find: ‘(standard input)’:1: invalid zero-length file name
foo
```

tavianator: Maybe you could use [split_inclusive()](https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.split_inclusive)?

Crypto-Darth: Surely.

src/find/mod.rs
```diff
@@ -138,6 +168,56 @@ fn parse_args(args: &[&str]) -> Result<ParsedInfo, Box<dyn Error>> {
     })
 }
 
+fn parse_files0_args(config: &Config, paths: &mut Vec<String>) -> Result<(), Box<dyn Error>> {
+    if config.from_file.as_deref() == Some("-") {
+        // small check if using stdin / pipe mode but data is not piped
+        if atty::is(Stream::Stdin) {
```
tavianator, line 174:
```suggestion
        if io::stdin().is_terminal() {
```

src/find/mod.rs
```diff
@@ -112,24 +116,50 @@ fn parse_args(args: &[&str]) -> Result<ParsedInfo, Box<dyn Error>> {
...
                 break;
             }
+            // https://www.gnu.org/software/findutils/manual/html_node/find_html/Starting-points.html
+            // This allows users to take the entry point for find from stdin (eg. pipe) or from a text file.
+            // eg. dummy | find -files0-from -
+            // eg. find -files0-from rust.txt -name "cargo"
+            "-files0-from" => {
+                // NOTE : -ok and -okdir should NOT be provided along with -files0-from args.
+                // This if checking -ok or -okdir exists in args should later be shifted to -ok and -okdir when it is fully implemented.
+                if args.iter().any(|arg| *arg == "-ok" || *arg == "-okdir") {
```
tavianator, line 126: `find -files0-from file -ok` is allowed, it's just `find -file0-from - -ok` that's forbidden

src/find/mod.rs
```diff
@@ -112,24 +116,50 @@ fn parse_args(args: &[&str]) -> Result<ParsedInfo, Box<dyn Error>> {
...
-        && (args[i] == "-" || !args[i].starts_with('-'))
-        && args[i] != "!"
-        && args[i] != "("
-    {
-        paths.push(args[i].to_string());
-        i += 1;
-    }
-    if i == paths_start {
-        paths.push(".".to_string());
+    // Check to see if '-files0-from' argument is used.
```
tavianator, line 144: `-files0-from` should be treated like a normal option, e.g. this should work

```console
$ find -follow -files0-from input
```

src/find/mod.rs
```diff
@@ -138,6 +168,56 @@ fn parse_args(args: &[&str]) -> Result<ParsedInfo, Box<dyn Error>> {
...
+        }
+
+        for value in file_split {
+            // empty starting point if detected shall make the program exit with non-zero code (as per GNU Manual)
+            // empty file should also exit immediately
+            // this if also handles if there are 2 consecutive ASCII NUL Characters
+            let path = std::str::from_utf8(value)?;
+            if path.is_empty() {
+                return Err(
+                    "Empty starting point detected in -files0-from input OR File is empty".into(),
```
tavianator, line 211: "File is empty" is not an error, and in fact if you pass an empty file you don't get an error message.

Also, an empty path should print an error but continue, like GNU find:

```console
$ printf '\0src\0' | find -files0-from -
find: ‘(standard input)’:1: invalid zero-length file name
src
src/find
...
```

Crypto-Darth: Yes sorry that's a mistake on my part , had fixed it but forgot to change the error message . Thank you

src/find/mod.rs
Crypto-Darth: Done.
Thank you

src/find/mod.rs
Crypto-Darth: Done.
Thank you

src/find/mod.rs
Crypto-Darth: Done. Thank you

src/find/mod.rs
Crypto-Darth: The issue with this is that `files0-from` can't be put into the matchers as it provides directories for walkdir.
Although the same command in the following order will work -> find -files0-from input -follow

I am a bit unsure on how to tackle this 
I will try to brainstorm but any ideas would be appreciated.

Crypto-Darth: Also just checked that even in gnu testsuite , all the test cases are using `find -files0-from.....` only
Therefore , this setting should also work for passing testcases.
Would like your opinion on this.

profile.json.gz
tertsdiepraam, line 1: I think you might have committed this by mistake? Could you remove it?

src/find/mod.rs
```diff
@@ -112,24 +116,51 @@ fn parse_args(args: &[&str]) -> Result<ParsedInfo, Box<dyn Error>> {
...
+                config.from_file = Some(args[i].to_string());
+
+                // NOTE : -ok and -okdir should NOT be provided along with -files0-from args.
+                // This if checking -ok or -okdir exists in args should later be shifted to -ok and -okdir when it is fully implemented.
+                if args.iter().any(|arg| *arg == "-ok" || *arg == "-okdir") && args[i] == "-" {
+                    return Err(From::from(
+                        "files0-from standard input cannot be combined with -ok, -okdir"
+                            .to_string(),
+                    ));
+                }
```
tertsdiepraam, line 137: Instead of doing this, you could add the parsing code for `-ok` and `-okdir` without implementing them and then emitting a warning at the end if both options have been set.

This version has to iterate over the arguments twice and also misses some important things like stopping at `--`.

tertsdiepraam (commented):
Sorry I can't help with the question you asked because I don't know the code well enough, but I have some other comments.

profile.json.gz
Crypto-Darth: Yes sorry , not sure where that crept in.
removed in latest commit

src/find/mod.rs
Crypto-Darth: Just to confirm , Parsing code means creating the parser file in `src/find/matchers` , and only implement the checker checking if `files0-from` was used alongside -ok & -okdir or not. Correct?

src/find/mod.rs
tavianator: I believe the goal of uutils is not just to pass the testsuites, but to act as a drop-in replacement.  Eventually, uutils find should be able to replace GNU find without breaking anyone's scripts, so it's important to aim for as much compatibility as possible.

What I'd recommend is changing `build_top_level_matcher()` and friends to take `&mut paths`, then parse `-files0-from` just like any other option.

src/find/mod.rs
tavianator: I think if you follow my suggestion here: https://github.com/uutils/findutils/pull/509#discussion_r2016767243, it will be easy to put this check directly in the parsing code for `-ok`/`-okdir`.

Right now this check is too aggressive, since it matches `-ok` even when it's not an option:

```console
$ printf 'src' | find -files0-from - -printf -ok
-ok-ok-ok-ok-ok-ok-ok-ok-ok-ok-ok-ok-ok-ok-ok
$ printf 'src' | ./target/debug/find -files0-from - -printf -ok
Error: files0-from standard input cannot be combined with -ok, -okdir
```

src/find/mod.rs
Crypto-Darth: Noted. will try to do the same as mentioned.

src/find/mod.rs
Crypto-Darth: I will follow your suggestion and try to complete the mentioned task.
thank you.

src/find/mod.rs
Crypto-Darth: image

src/find/mod.rs
Crypto-Darth: image
Working as intended

src/find/mod.rs
Crypto-Darth: image

src/find/mod.rs
Crypto-Darth: image
Now working as intended.

Crypto-Darth:
@tavianator 
I have transformed -files0-from to work and act as an matcher / normal argument.

I believe I found a better solution by storing a `new_paths` in the `Config` Struct rather than passing `&mut paths` and the program works as intended.
Let me know if this approach is acceptable or not.

Thank you.

src/find/mod.rs
Crypto-Darth: @tertsdiepraam I believe the latest commit solves this and now isn't an aggressive check nor double iteration when matching `args`.

src/find/matchers/mod.rs
```diff
@@ -892,6 +894,30 @@ fn build_matcher_tree(
                 config.version_requested = true;
                 None
             }
+            "-files0-from" => {
+                // The logic for the below if is to handle edge cases like `-files0-from -ok` etc.
```
tavianator, line 898: I don't think you should do this.  GNU find allows files that start with a dash:

```console
tavianator@graphene $ printf src > ./-ok
tavianator@graphene $ find -files0-from -ok
src
src/find
...
```

src/find/matchers/mod.rs
```diff
@@ -67,6 +58,17 @@ use self::time::{
 };
 use self::type_matcher::{TypeMatcher, XtypeMatcher};
 use self::user::{NoUserMatcher, UserMatcher};
+use ::regex::Regex;
```
tavianator, line 61: Can you keep the old grouping?

src/find/matchers/mod.rs
```diff
@@ -952,6 +978,69 @@ fn build_matcher_tree(
...
 }
 
+// https://www.gnu.org/software/findutils/manual/html_node/find_html/Starting-points.html
+// This allows users to take the entry point for find from stdin (eg. pipe) or from a text file.
+// eg. dummy | find -files0-from -
+// eg. find -files0-from rust.txt -name "cargo"
+fn parse_files0_args(config: &mut Config, args: &[&str]) -> Result<(), Box<dyn Error>> {
+    let mode = args
+        .iter()
+        .position(|m| *m == "-files0-from")
```
tavianator, line 988: Why not just pass `i` to `parse_files0_args`, instead of scanning the whole argument list again?

src/find/matchers/mod.rs
```diff
@@ -952,6 +978,69 @@ fn build_matcher_tree(
...
+        .expect("Error getting files0-from mode.");
+
+    if args[mode + 1] == "-" {
+        // small check if using stdin / pipe mode but data is not piped
+        if io::stdin().is_terminal() {
+            return Err(From::from("stdin not piped".to_string()));
```
tavianator, line 994: GNU find doesn't do this check

src/find/matchers/mod.rs
```diff
@@ -952,6 +978,69 @@ fn build_matcher_tree(
...
+        if string_segments.iter().any(|s| s.is_empty()) {
+            println!("Find : Empty starting point detected in -files0-from input");
+            //remove the empty ones so as to avoid file not found error
+            string_segments.retain(|s| !s.is_empty());
+        }
+        config
+            .new_paths
+            .get_or_insert(Vec::new())
+            .extend(string_segments);
+    } else {
```
tavianator, line 1019: Why duplicate so much between these branches?  It could be more like

```rust
let mut buffer = vec![];
if args[i + 1] == "-" {
    stdin().read_to_end(&mut buffer)?;
} else {
    let mut file = File::open(args[i + 1])?;
    file.read_to_end(&mut buffer)?;
}

let mut split = buffer.split(...
```

src/find/matchers/mod.rs
```diff
@@ -892,6 +894,30 @@ fn build_matcher_tree(
...
+                    return Err(From::from(format!("missing argument to {}", args[i])));
+                }
+                parse_files0_args(config, args)?;
+                i += 1;
+                None
+            }
+            // -ok and -okdir are yet to be implemented fully
+            "-ok" | "-okdir" => {
+                //basic check for if -files0-from and -ok /-okdir are used together
+                if let Some(pos) = args.iter().position(|arg| *arg == "-files0-from") {
```
tavianator, line 911: This is too aggressive, it will match `-files0-from` even when it's not an option, e.g.

```console
$ find -fprintf -files0-from - -ok echo {} \;
```

Instead you should have the `-files0-from` parser set a flag on `config` to say that stdin is already used, and check that when parsing `-ok`.

You also have to do it the other way around, to handle

```console
$ find -ok echo {} \; -files0-from -
find: warning: you have specified the global option -files0-from after the argument -ok, but global options are not positional, i.e., -files0-from affects tests specified before it as well as those specified after it.  Please specify global options before other arguments.
find: option -files0-from reading from standard input cannot be combined with -ok, -okdir
```

tavianator (changes requested):

src/find/matchers/mod.rs
Crypto-Darth: fixed. thank you

src/find/matchers/mod.rs
Crypto-Darth: I'm not sure what you exactly meant by grouping but if it's about having the imports arranged then I have done that in the latest commits.

src/find/matchers/mod.rs
Crypto-Darth: Indeed.
Fixed it . thank you.

src/find/matchers/mod.rs
Crypto-Darth: removed.

src/find/matchers/mod.rs
Crypto-Darth: yes much better approach :)
have updated the logic in latest commits.
thank you

src/find/matchers/mod.rs
Crypto-Darth: I have setup the flag and used that for checking in -ok / -okdir but I am a bit clueless on how to get the desired output for ->
```
find -fprintf -files0-from - -ok echo {} \;
```
and how to setup this warning message 
```
find: warning: you have specified the global option -files0-from after the argument -ok, but global options are not positional, i.e., -files0-from affects tests specified before it as well as those specified after it.  Please specify global options before other arguments.
``` 
as shown by you in your previous message without an aggressive check.
Any ideas would be appreciated.

src/find/matchers/mod.rs
tavianator: Yeah, I mean having the original groups of imports in the same place they started

src/find/matchers/mod.rs
tavianator: Oh you don't need to add the warning.

src/find/matchers/mod.rs
tavianator: The comment needs updating

src/find/matchers/mod.rs
```diff
@@ -892,6 +895,26 @@ fn build_matcher_tree(
...
                 None
             }
+            "-files0-from" => {
+                // The logic for the below if is to handle edge cases like `-files0-from -ok` etc.
+                if i >= args.len() - 1 {
+                    return Err(From::from(format!("missing argument to {}", args[i])));
+                }
+                parse_files0_args(config, args[i + 1])?;
+                i += 1;
+                None
```
tavianator, line 905:
```suggestion
                Some(TrueMatcher.into_box())
```

src/find/matchers/mod.rs
```diff
@@ -892,6 +895,26 @@ fn build_matcher_tree(
...
+            }
+            // -ok and -okdir are yet to be implemented fully
```
tavianator, line 907: Sorry, I didn't realize these weren't even implemented originally.  You might as well just keep them not implemented, so every use of `-ok` will error.  Then it will be up to whoever implements `-ok` to make the interaction with `-files0-from` work ;)

tavianator (changes requested):

src/find/matchers/mod.rs
Crypto-Darth: oh alright then

src/find/matchers/mod.rs
Crypto-Darth: done

Crypto-Darth:
Regarding GNU Test-suite @tavianator ,

 I have manually tested each testcases in the files0-from test of GNU.
 Here's my findings->
 
 There are basically 2 problems , one is a feature not implemented and other is working output but different error message which would make the test case fail.
 
 I have compiled all into [this file](https://gist.github.com/Crypto-Darth/81834e37f9da6c2dbb2d6a39ca37ce4a) here which can be used as reference.
 
 I'm a bit unsure if the testcases failing between 2-8 is because of `-ok/-okdir` not fully implemented or not .
 
 Would like your opinion on this.

tavianator:
Right, I think you can ignore the ones with `-ok` for now.  Test 2 should be fixed here.  And I'm not sure whether this project aims for identical error messages to GNU find, but if it helps pass test cases you might as well re-word the error messages if you can.

Crypto-Darth:
> Test 2 should be fixed here. 

Regarding this , was thinking of adding an if in the `parse_args` function which checks if `is_stdin = true` or not but the problem is that `is_stdin` is set to true later , so either check `if -files0-from in parse_args` and only set `is_stdin to true` and then below check `if is_stdin true and return error` so as to disallow file operands.
let me know if this works.

> but if it helps pass test cases you might as well re-word the error messages if you can.

On it . I'll make the message similar enough to pass their regex.


@tavianator

Crypto-Darth:
@tavianator I'm still clueless on how should I go on to fix testcase 2 ,
Would request your suggestion.

src/find/matchers/mod.rs
tavianator: You marked this as resolved but the imports are still in a different place.  I was suggesting to keep them above, where they used to be.

src/find/matchers/mod.rs
```diff
@@ -817,7 +820,9 @@ fn build_matcher_tree(
             }
             ")" => {
                 if !expecting_bracket {
-                    return Err(From::from("you have too many ')'"));
+                    return Err(From::from(
+                        " invalid expression: expected expression before closing parentheses ')'.",
```
tavianator, line 824:
```suggestion
                        "invalid expression: expected expression before closing parentheses ')'.",
```

src/find/matchers/mod.rs
```diff
@@ -892,6 +897,25 @@ fn build_matcher_tree(
...
+            "-ok" | "-okdir" => {
+                //basic check for if -files0-from and -ok /-okdir are used together
+                if config.is_stdin {
+                    return Err(From::from(
+                        "files0-from standard input cannot be combined with -ok / -okdir"
+                            .to_string(),
+                    ));
+                }
+                None
+            }
```
tavianator, line 918:
```suggestion
```

Crypto-Darth: May i ask why we removed the -ok and -okdir parsing code? @tavianator

tavianator: https://github.com/uutils/findutils/pull/509#discussion_r2020278677

Crypto-Darth: oh alright then , ill remove the respective test case too.

Crypto-Darth:
> So looking more at how GNU find works here, I think what you should do is: when you see `-files0-from`, save the argument in the `Config`, but _don't open/read it yet_. Then, after parsing everything, if you saved a `-files0-from` argument, go read it and add all the root paths. At that point you can also check if any other paths were already given, and give an error instead.

@tavianator , 
I think I found a better way by writing to the new paths directly via insert() rather than get_and_insert().
Therefore if there's more than 1 usage of -files0-from the last one will be considered.

image


---

Regarding the original Issue (Test 2) is now also solved.

image


Please review the latest commit .
Thank you.

tavianator:
> @tavianator ,
> I think I found a better way by writing to the new paths directly via insert() rather than get_and_insert().
> Therefore if there's more than 1 usage of -files0-from the last one will be considered.

That would almost work, but you also have to ensure that

```console
$ find -files0-from - -files0-from file
```

doesn't read from stdin at all, so I think it's easiest to delay reading the file.

Crypto-Darth:
> > @tavianator ,
> > I think I found a better way by writing to the new paths directly via insert() rather than get_and_insert().
> > Therefore if there's more than 1 usage of -files0-from the last one will be considered.
> 
> That would almost work, but you also have to ensure that
> 
> ```
> $ find -files0-from - -files0-from file
> ```
> 
> doesn't read from stdin at all, so I think it's easiest to delay reading the file.

---

To be really honest that is a very edge case , I don't see any user or a script using -files0-from twice but I understand the need for maximum compatibility hence, it has been fixed as per your suggestion in latest commit.

image

src/find/matchers/mod.rs
```diff
@@ -952,6 +968,45 @@ fn build_matcher_tree(
...
+    }
+
+    let mut string_segments: Vec<String> = buffer_split
+        .iter()
+        .filter_map(|s| std::str::from_utf8(s).ok())
+        .map(|s| s.to_string())
+        .collect();
+    // empty starting point checker
+    if string_segments.iter().any(|s| s.is_empty()) {
+        println!("find: invalid zero-length file name");
```
tavianator, line 1001: This should be `eprintln!()`, and you should also set the exit code to 1

Crypto-Darth: Any clue on how can I set the exit code to 1?
I don't see how current implementation allows me to set custom exit code without exiting in between of execution.

tavianator: Seems like it would take more refactoring to make that work nicely.  Feel free to defer it to a future PR

src/find/matchers/mod.rs
```diff
@@ -943,6 +956,9 @@ fn build_matcher_tree(
             }
         }
     }
+    if config.files0_argument.is_some() {
```
tavianator, line 959: This should probably go after the next if

src/find/matchers/mod.rs
```diff
@@ -952,6 +968,45 @@ fn build_matcher_tree(
...
+    if mode == "-" {
+        std::io::stdin().read_to_end(&mut buffer)?;
+    } else {
+        let mut file =
+            File::open(mode).map_err(|e| format!("cannot open '{}' for reading: {}", mode, e))?;
+        file.read_to_end(&mut buffer)?;
+    }
+
+    let mut buffer_split: Vec<&[u8]> = buffer.split(|&b| b == 0).collect();
+    //if the pipe/file ends with ASCII NULL
```
tavianator, line 989:
```suggestion
    // if the pipe/file ends with ASCII NULL
```

src/find/matchers/mod.rs
```diff
@@ -952,6 +968,45 @@ fn build_matcher_tree(
...
+
+    let mut string_segments: Vec<String> = buffer_split
+        .iter()
+        .filter_map(|s| std::str::from_utf8(s).ok())
+        .map(|s| s.to_string())
+        .collect();
+    // empty starting point checker
+    if string_segments.iter().any(|s| s.is_empty()) {
+        eprintln!("find: invalid zero-length file name");
+        //remove the empty ones so as to avoid file not found error
```
tavianator, line 1002:
```suggestion
        // remove the empty ones so as to avoid file not found error
```

src/find/mod.rs
```diff
@@ -202,6 +216,7 @@ fn do_find(args: &[&str], deps: &dyn Dependencies) -> Result<i32, Box<dyn Error>
 
     let mut ret = 0;
     let mut quit = false;
+
```
tavianator, line 219:
```suggestion
```

src/find/matchers/mod.rs
Crypto-Darth: Surely.

src/find/matchers/mod.rs
Crypto-Darth: Fixed in latest commit.

tavianator:
This LGTM now except for those few nits.  Sorry for so much back and forth, it's a surprisingly finicky feature!

Crypto-Darth:
> This LGTM now except for those few nits. Sorry for so much back and forth, it's a surprisingly finicky feature!

No worries :)
Really got to learn a lot.
Thanks a lot to you for your time and patience :)
Thanks to @sylvestre and @tertsdiepraam too for their time and reviews.

Crypto-Darth:
@hanbings @sylvestre ,
Can we get this merged now if all good to go ? :)

hanbings:
This is truly amazing work! Many thanks to all the contributors!

Issue #439, opened 2024-08-11
Specifying the same file for output multiple times in find will overwrite the file by default.


hanbings:
refer: https://github.com/uutils/findutils/pull/421#discussion_r1705640231

> For future reference: GNU find de-duplicates files, so if you do find -fprint foo -fprint foo it will use the same FILE * for both and you'll get each line twice. Right now uutils will use two different Files and the writes will overlap with each other.

tonakai-s:
Hello!
Is someone working on it? I would like to give it a try, if not.

hanbings:
> Is someone working on it? I would like to give it a try, if not.

As far as I know, no yet. Contributions are welcome! :)

PR #511, opened 2025-03-24
bugfix: Reuse the same file pointer for repeated `find` output (#439)

tonakai-s:
This PR creates a kind of memoization to the file creation inside the findutils/find command.
Uses a hashmap to know which file paths already has been created and returns the Arc<File> to prevent overwrite.

It also add a new testing function `find_using_same_out_multiple_times` to these more "complex" cases.
But, the expected outputs are fixed, if the assert would be better by catching the expected value dynamically in case of the test_data/simple structure change, let me know!

### DIFF
--- a/src/find/matchers/ls.rs
+++ b/src/find/matchers/ls.rs
@@ -7,6 +7,7 @@ use chrono::DateTime;
 use std::{
     fs::File,
     io::{stderr, Write},
+    sync::Arc,
 };
 
 use super::{Matcher, MatcherIO, WalkEntry};
@@ -110,11 +111,11 @@ fn format_permissions(file_attributes: u32) -> String {
 }
 
 pub struct Ls {
-    output_file: Option<File>,
+    output_file: Option<Arc<File>>,
 }
 
 impl Ls {
-    pub fn new(output_file: Option<File>) -> Self {
+    pub fn new(output_file: Option<Arc<File>>) -> Self {
         Self { output_file }
     }
 
@@ -256,7 +257,7 @@ impl Ls {
 impl Matcher for Ls {
     fn matches(&self, file_info: &WalkEntry, matcher_io: &mut MatcherIO) -> bool {
         if let Some(file) = &self.output_file {
-            self.print(file_info, file, true);
+            self.print(file_info, file.clone(), true);
         } else {
             self.print(
                 file_info,
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -35,8 +35,10 @@ use ::regex::Regex;
 use chrono::{DateTime, Datelike, NaiveDateTime, Utc};
 use fs::FileSystemMatcher;
 use ls::Ls;
+use std::collections::HashMap;
 use std::fs::{File, Metadata};
 use std::path::Path;
+use std::sync::Arc;
 use std::time::SystemTime;
 use std::{error::Error, str::FromStr};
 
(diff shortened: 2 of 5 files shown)


hanbings:
Hi, I see some tests failed, could you please check them? Thanks.

tonakai-s:
I executed the test suite on my windows machine and it failed, and I don't know why certainly.
I will run the `fmt` and check for solutions...
(Now I understand why other tests not checked all items, sorry about the waste of time)

src/find/matchers/mod.rs
```diff
@@ -268,13 +270,34 @@ impl ComparableValue {
     }
 }
 
+// Used on file output arguments.
+// When the same file is specified multiple times, the same pointer is used.
+struct FileMemoizer {
+    mem: HashMap<String, Arc<File>>,
```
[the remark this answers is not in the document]
tonakai-s: True, newbie mistake of my part, will fix it.

src/find/matchers/ls.rs
```diff
@@ -256,7 +257,7 @@ impl Ls {
 impl Matcher for Ls {
     fn matches(&self, file_info: &WalkEntry, matcher_io: &mut MatcherIO) -> bool {
         if let Some(file) = &self.output_file {
-            self.print(file_info, file, true);
+            self.print(file_info, file.clone(), true);
```
tavianator, line 260:
```suggestion
            self.print(file_info, file.as_ref(), true);
```

Same everywhere else, there's usually no need to clone the `Arc`

tests/find_cmd_tests.rs
```diff
@@ -12,8 +12,10 @@ use assert_cmd::Command;
 use predicates::prelude::*;
 use regex::Regex;
 use serial_test::serial;
+use std::collections::HashMap;
 use std::fs::{self, File};
 use std::io::{Read, Write};
+use std::os::unix;
```
hanbings, line 18: This has caused a test failure on the Windows platform. Could you take a look?
Operating system-specific `use` should be wrapped in `#[cfg(target_os)]`.

src/find/matchers/mod.rs
```diff
@@ -268,13 +271,54 @@ impl ComparableValue {
...
+
+        let file = self
+            .mem
+            .entry(file_id.unwrap())
+            .or_insert(Arc::new(File::create(path)?));
+        Ok(file.clone())
+    }
+
+    fn get_file_id(&self, path: &Path) -> Result<u64, Box<dyn Error>> {
+        let file_info = FileInformation::from_path(&path, true)?;
```
hanbings, line 303: And here's a clippy.

```
error: the borrowed expression implements the required traits
   --> src/find/matchers/mod.rs:303:52
    |
303 |         let file_info = FileInformation::from_path(&path, true)?;
    |                                                    ^^^^^ help: change this to: `path`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrows_for_generic_args
    = note: `-D clippy::needless-borrows-for-generic-args` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::needless_borrows_for_generic_args)]`

```

hanbings (changes requested):

src/find/matchers/mod.rs
```diff
@@ -268,13 +271,54 @@ impl ComparableValue {
     }
 }
 
+// Used on file output arguments.
+// Based on path inode or file_index check if the file already has been specified.
+// If yes, use the same file pointer.
+struct FileMemoizer {
+    mem: HashMap<u64, Arc<File>>,
```
tavianator, line 278: A single `u64` is not enough to uniquely identify a file.  On UNIX-like systems, you need the pair `(st_dev, st_ino)` (AKA device ID + inode number) at least for a unique ID.  Not sure what's necessary on Windows.  But using just the inode number is wrong since inode numbers are only unique within a single file system.

tonakai-s: Sure, I read about it sometime ago but forgot, thanks for the precise explanation.
I will use the `std::fs::metadata` to take the st_dev and st_ino from the same source, but using `std::os::unix::fs::MetadataExt` and `std::os::windows::fs::MetadataExt`.
Will do a manual test to check the assurance of this implementation, but I don't think that has a way to do an automated test of it, what do you think?

tavianator: Maybe use uucore's FileInformation?  Check samefile.rs for an example.  (Hopefully it impls Hash)

sylvestre:
it needs to be rebased, sorry!

sylvestre:
Did you run some benchmarks in this? Thanks

tonakai-s:
> Did you run some benchmarks in this? Thanks

Not yet, I can run if needed, do you guys have a pattern?

PR #515, opened 2025-03-27
find: Implement exec[dir] +

milkory:
This PR implements `-exec ... {} +` with [argmax](https://crates.io/crates/argmax) as a new dependency.

Solve #6 , #366

### DIFF
--- a/src/find/matchers/exec.rs
+++ b/src/find/matchers/exec.rs
@@ -4,6 +4,7 @@
 // license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
+use std::cell::RefCell;
 use std::error::Error;
 use std::ffi::OsString;
 use std::io::{stderr, Write};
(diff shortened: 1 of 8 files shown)


src/find/matchers/exec.rs
```diff
@@ -97,6 +98,108 @@ impl Matcher for SingleExecMatcher {
...
+                    Some(parent) if parent == Path::new("") => {
+                        // Paths like "foo" have a parent of "".  Avoid chdir("").
+                    }
+                    Some(parent) => {
+                        command.current_dir(parent);
+                    }
+                }
+            }
+            if let Err(e) = command.status() {
+                writeln!(&mut stderr(), "Failed to run {}: {}", self.executable, e).unwrap();
```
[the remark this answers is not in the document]
milkory: Command execution may also fail at `finished_dir` and `finished`, should I add `matcher_io` to them?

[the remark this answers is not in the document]
tavianator: Yeah, any failed exec+ command should set an error code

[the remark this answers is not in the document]
milkory: Added, thanks.

src/find/mod.rs
```diff
@@ -170,6 +172,14 @@ fn process_dir(
             Ok(entry) => {
                 let mut matcher_io = matchers::MatcherIO::new(deps);
 
+                let new_dir = entry.path().parent().map(|x| x.to_path_buf());
+                if new_dir != current_dir {
+                    if let Some(dir) = current_dir.take() {
+                        matcher.finished_dir(dir.as_path(), &mut matcher_io);
+                    }
+                    current_dir = new_dir
```
hanbings, line 180: Hi, looks like there is some clippy issue here.
Could you please fix it? Thanks.

milkory: Fixed, thanks.

src/find/matchers/exec.rs
```diff
@@ -97,6 +98,116 @@ impl Matcher for SingleExecMatcher {
...
+                    Some(parent) => {
+                        command.current_dir(parent);
+                    }
+                }
+            }
+            self.run_command(command, matcher_io);
+
+            // Reset command status.
+            *command = self.new_command();
+            command.try_arg(&path_to_file).unwrap();
```
tavianator, line 180: I think it's possible for this to fail for particularly long file paths.  We should report an error rather than panic if that happens

src/find/matchers/exec.rs
```diff
@@ -97,6 +98,116 @@ impl Matcher for SingleExecMatcher {
     }
 }
 
+pub struct MultiExecMatcher {
+    executable: String,
+    args: Vec<OsString>,
+    exec_in_parent_dir: bool,
+    /// Command to build while matching.
+    command: RefCell<Option<argmax::Command>>,
```
tavianator, line 106: Why does this need to be an Option?

src/find/matchers/mod.rs
```diff
@@ -1519,6 +1539,30 @@ mod tests {
         } else {
             panic!("parsing argument list with exec and no executable should fail");
         }
+
+        if let Err(e) = build_top_level_matcher(&["-exec", "+"], &mut config) {
+            assert!(e.to_string().contains("missing argument"));
+        } else {
+            panic!("parsing argument list with exec and no executable should fail");
```
tavianator, line 1546: Can you use expect_err() for these?

src/find/matchers/exec.rs
milkory: Option is used to prevent running empty command when nothing matched.

tavianator: Ah okay that makes sense

src/find/matchers/mod.rs
milkory: They are just copied from other tests. And use expect_err will cause `dyn find::matchers::Matcher doesn't implement Debug`. I don't know how to handle it.

hanbings (approved):
lgtm, will there be any other updates here? : )

src/find/matchers/exec.rs
milkory: I've added a check on that, thanks.

milkory:
I don't have more updates if no further issue.

PR #522, opened 2025-04-03
clippy: fix warnings from Rust 1.86

cakebaker:
This PR fixes warnings from the [manual_div_ceil](https://rust-lang.github.io/rust-clippy/master/index.html#manual_div_ceil) and [double_ended_iterator_last](https://rust-lang.github.io/rust-clippy/master/index.html#double_ended_iterator_last) lints.

### DIFF
--- a/src/find/matchers/printf.rs
+++ b/src/find/matchers/printf.rs
@@ -401,7 +401,7 @@ fn format_directive<'entry>(
             // even though it's arguably not 100% correct.
             if *large_blocks {
                 // Ceiling divide in half.
-                (blocks + 1) / 2
+                blocks.div_ceil(2)
             } else {
                 blocks
             }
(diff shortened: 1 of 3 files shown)


src/find/matchers/printf.rs
```diff
@@ -401,7 +401,7 @@ fn format_directive<'entry>(
             // even though it's arguably not 100% correct.
             if *large_blocks {
                 // Ceiling divide in half.
-                (blocks + 1) / 2
+                blocks.div_ceil(2)
```
tavianator, line 404: Might as well make the same fix right above: https://github.com/uutils/findutils/blob/5235d782916f51b31ba9e1f41fae717ba6dfd871/src/find/matchers/printf.rs#L396

Side note: I think the code here is wrong, but that can be fixed in a different PR.  It should be more like

```rust
let blocks = meta()?.blocks();
let len = blocks * STANDARD_BLOCK_SIZE;
let bs = if *large_blocks { 1024 } else { 512 };
blocks.div_ceil(len, bs)
```

PR #527, opened 2025-04-05
Implement : Chained Type matcher arguments

Crypto-Darth:
Refer - https://www.gnu.org/software/findutils/manual/html_node/find_html/Type.html

>As a GNU extension, multiple file types can be provided as a combined list separated by comma ‘,’. For example, ‘-type f,d,l’ is logically interpreted as ‘( -type f -o -type d -o -type l )’.

### DIFF
--- a/src/find/matchers/type_matcher.rs
+++ b/src/find/matchers/type_matcher.rs
@@ -62,9 +62,7 @@ impl TypeMatcher {
 
 impl Matcher for TypeMatcher {
     fn matches(&self, file_info: &WalkEntry, _: &mut MatcherIO) -> bool {
-        self.file_type
-            .iter()
-            .any(|entry| *entry == file_info.file_type())
+        self.file_type.contains(&file_info.file_type())
     }
 }
 
--- a/src/find/matchers/type_matcher.rs
+++ b/src/find/matchers/type_matcher.rs
@@ -91,22 +91,17 @@ impl Matcher for XtypeMatcher {
 
         let file_type_result = follow
             .metadata(file_info)
-            .map(|m| m.file_type())
-            .map(FileType::from);
+            .map(|m| m.file_type().into())
+            .or_else(|e| {
+                if e.is_loop() {
+                    Ok(FileType::Symlink)
+                } else {
+                    Err(e)
+                }
+            })
+            .unwrap_or(FileType::Unknown);
 
-        for expected_type in &self.file_type {
-            if let Ok(file_type) = file_type_result {
-                if file_type == *expected_type {
-                    return true;
-                }
-            } else if let Err(e) = &file_type_result {
-                // Since GNU find 4.10, ELOOP will match -xtype l
-                if e.is_loop() && *expected_type == FileType::Symlink {
-                    return true;
-                }
-            }
-        }
-        false
+        self.file_type.contains(&file_type_result)
     }
 }
 
--- a/src/find/matchers/type_matcher.rs
+++ b/src/find/matchers/type_matcher.rs
@@ -4,15 +4,13 @@
 // license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
-use std::error::Error;
-
 use super::{FileType, Follow, Matcher, MatcherIO, WalkEntry};
-
-type TypeList = std::collections::HashSet<FileType>;
+use std::collections::HashSet;
+use std::error::Error;
 
 /// This matcher checks the type of the file.
 pub struct TypeMatcher {
-    file_type: TypeList,
+    file_type: HashSet<FileType>,
 }
 
 fn parse(type_string: &str, mode: &str) -> Result<FileType, Box<dyn Error>> {
@@ -68,7 +66,7 @@ impl Matcher for TypeMatcher {
 
 /// Like [TypeMatcher], but toggles whether symlinks are followed.
 pub struct XtypeMatcher {
-    file_type: TypeList,
+    file_type: HashSet<FileType>,
 }
 
 impl XtypeMatcher {
@@ -103,7 +101,7 @@ impl Matcher for XtypeMatcher {
     }
 }
 
-fn type_creator(type_string: &str, mode: &str) -> Result<TypeList, Box<dyn Error>> {
+fn type_creator(type_string: &str, mode: &str) -> Result<HashSet<FileType>, Box<dyn Error>> {
     let mut file_types = std::collections::HashSet::new();
 
     if type_string.contains(',') {
--- a/src/find/matchers/type_matcher.rs
+++ b/src/find/matchers/type_matcher.rs
@@ -341,8 +341,46 @@ mod tests {
         assert!(TypeMatcher::new("").is_err());
         assert!(TypeMatcher::new("f,f").is_err());
         assert!(TypeMatcher::new("f,").is_err());
+        assert!(TypeMatcher::new("x,y").is_err());
+        assert!(TypeMatcher::new("fd").is_err());
+
         assert!(XtypeMatcher::new("").is_err());
         assert!(XtypeMatcher::new("f,f").is_err());
         assert!(XtypeMatcher::new("f,").is_err());
+        assert!(XtypeMatcher::new("x,y").is_err());
+        assert!(XtypeMatcher::new("fd").is_err());
+    }
+
+    #[test]
+    fn type_matcher_multiple_valid_types() {
+        let deps = FakeDependencies::new();
+        let file = get_dir_entry_for("test_data/simple", "abbbc");
+        let dir = get_dir_entry_for("test_data", "simple");
+        let symlink = get_dir_entry_for("test_data/links", "link-f");
+
+        let matcher = TypeMatcher::new("f,d").unwrap();
+        assert!(matcher.matches(&file, &mut deps.new_matcher_io()));
+        assert!(matcher.matches(&dir, &mut deps.new_matcher_io()));
+        assert!(!matcher.matches(&symlink, &mut deps.new_matcher_io()));
+
+        let matcher = TypeMatcher::new("l,d").unwrap();
+        assert!(!matcher.matches(&file, &mut deps.new_matcher_io()));
+        assert!(matcher.matches(&dir, &mut deps.new_matcher_io()));
+        assert!(matcher.matches(&symlink, &mut deps.new_matcher_io()));
+    }
+
+    #[cfg(unix)]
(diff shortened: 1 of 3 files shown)


src/find/matchers/type_matcher.rs
```diff
@@ -39,29 +45,101 @@ fn parse(type_string: &str) -> Result<FileType, Box<dyn Error>> {
...
     pub fn new(type_string: &str) -> Result<Self, Box<dyn Error>> {
-        let file_type = parse(type_string)?;
-        Ok(Self { file_type })
+        let mut single_file_type: Option<FileType> = None;
+        let mut chained_type_list: Option<Vec<FileType>> = None;
+        if type_string.contains(',') {
+            let mut seen = std::collections::HashSet::new();
+
+            chained_type_list = Some(
+                type_string
```
sylvestre, line 113: this is duplicated code from line 55
maybe move this into a function

src/find/matchers/type_matcher.rs
```diff
@@ -70,16 +148,35 @@ impl Matcher for XtypeMatcher {
...
             .map(FileType::from);
 
-        match file_type {
-            Ok(file_type) if file_type == self.file_type => true,
-            // Since GNU find 4.10, ELOOP will match -xtype l
-            Err(e) if self.file_type.is_symlink() && e.is_loop() => true,
-            _ => false,
+        if let Some(types) = &self.chained_file_types {
+            for expected_type in types {
+                match &file_type_result {
```
sylvestre, line 158: i think you can use an if / else here

src/find/matchers/type_matcher.rs
```diff
@@ -70,16 +148,35 @@ impl Matcher for XtypeMatcher {
...
+                            return true;
+                        }
+                    }
+                }
+            }
+            false
+        } else {
+            // Single type check
+            let expected_type = self.file_type.unwrap(); // Safe due to struct invariants
+            match file_type_result {
```
sylvestre, line 176: same here

src/find/matchers/type_matcher.rs
Crypto-Darth: Done :)

src/find/matchers/type_matcher.rs
Crypto-Darth: Done with latest commit

Crypto-Darth:
image

New testcase passing :)

Not sure why CI isn't sending message regarding it.

image
for comparison - above image is from head commit of main repo

src/find/matchers/type_matcher.rs
```diff
@@ -8,12 +8,16 @@ use std::error::Error;
...
 use super::{FileType, Follow, Matcher, MatcherIO, WalkEntry};
 
+type SingleTypeList = Option<FileType>;
+type ChainedTypeList = Option<Vec<FileType>>;
+
 /// This matcher checks the type of the file.
 pub struct TypeMatcher {
-    file_type: FileType,
+    file_type: SingleTypeList,
+    chained_file_types: ChainedTypeList,
```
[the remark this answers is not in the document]
Crypto-Darth: yep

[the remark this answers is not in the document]
Crypto-Darth: want me to use only 1 type for all?

sylvestre:
Could you please add tests at the binary level?
Not only at the function level :)

Merci

src/find/matchers/type_matcher.rs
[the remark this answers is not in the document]
Crypto-Darth: fixed

src/find/matchers/type_matcher.rs
```diff
@@ -8,12 +8,15 @@ use std::error::Error;
 
 use super::{FileType, Follow, Matcher, MatcherIO, WalkEntry};
 
+type TypeList = Option<Vec<FileType>>;
+
 /// This matcher checks the type of the file.
 pub struct TypeMatcher {
-    file_type: FileType,
+    file_type: TypeList,
+    chained_file_types: TypeList,
```
tavianator, line 16: I think there's still no reason to separate these two fields.  Just use a `Vec<FileType>` IMO.

Crypto-Darth:
> Could you please add tests at the binary level? Not only at the function level :)
> 
> Merci

@sylvestre Done :)

src/find/matchers/type_matcher.rs
Crypto-Darth: Done with latest commit

src/find/matchers/type_matcher.rs
```diff
@@ -70,17 +89,55 @@ impl Matcher for XtypeMatcher {
...
-            Ok(file_type) if file_type == self.file_type => true,
-            // Since GNU find 4.10, ELOOP will match -xtype l
-            Err(e) if self.file_type.is_symlink() && e.is_loop() => true,
-            _ => false,
+        for expected_type in &self.file_type {
+            if let Ok(file_type) = file_type_result {
+                if file_type == *expected_type {
+                    return true;
+                }
+            } else if let Err(e) = &file_type_result {
```
tavianator, line 102: Rather than do this on every iteration, you could just do

```rust
let file_type = follow
    .metadata(file_info)
    .map(|m| m.file_type().into())
    .or_else(|e| if e.is_loop() { Ok(FileType::Symlink) } else { e })
    .unwrap_or(FileType::Unknown);

self.file_types.contains(file_type)
```

src/find/matchers/type_matcher.rs
```diff
@@ -70,17 +89,55 @@ impl Matcher for XtypeMatcher {
...
+                }
+            }
+        }
+        false
+    }
+}
+
+fn type_creator(type_string: &str, mode: &str) -> Result<TypeList, Box<dyn Error>> {
+    if type_string.contains(',') {
+        let mut seen = std::collections::HashSet::new();
```
tavianator, line 115: Hmm, if you're building a `HashSet` anyway, why not store a `HashSet` instead of a `Vec`?

src/find/matchers/type_matcher.rs
Crypto-Darth: yes that's much better thank you
Done with latest commit.

src/find/matchers/type_matcher.rs
Crypto-Darth: Indeed a better approach , Thank you.
Done with latest Commit

src/find/matchers/type_matcher.rs
```diff
@@ -39,29 +53,34 @@ fn parse(type_string: &str) -> Result<FileType, Box<dyn Error>> {
...
+        })
     }
 }
 
 impl Matcher for TypeMatcher {
     fn matches(&self, file_info: &WalkEntry, _: &mut MatcherIO) -> bool {
-        file_info.file_type() == self.file_type
+        self.file_type
+            .iter()
+            .any(|entry| *entry == file_info.file_type())
```
tavianator, line 67:
```suggestion
        self.file_type.contains(&file_info.file_type())
```

src/find/matchers/type_matcher.rs
```diff
@@ -70,18 +89,48 @@ impl Matcher for XtypeMatcher {
             Follow::Always
         };
 
-        let file_type = follow
+        let file_type_result = follow
```
tavianator, line 92: This isn't a `Result` any more, might as well keep the old name

src/find/matchers/type_matcher.rs
```diff
@@ -70,18 +89,48 @@ impl Matcher for XtypeMatcher {
...
+                }
+            })
+            .unwrap_or(FileType::Unknown);
+
+        self.file_type.contains(&file_type_result)
+    }
+}
+
+fn type_creator(type_string: &str, mode: &str) -> Result<TypeList, Box<dyn Error>> {
+    let mut file_types = std::collections::HashSet::new();
```
tavianator, line 109: `use std::collections::HashSet`?

src/find/matchers/type_matcher.rs
```diff
@@ -8,12 +8,14 @@ use std::error::Error;
 
 use super::{FileType, Follow, Matcher, MatcherIO, WalkEntry};
 
+type TypeList = std::collections::HashSet<FileType>;
```
tavianator, line 11: I'm not sure this alias provides much value over just `HashSet<FileType>`

src/find/matchers/type_matcher.rs
Crypto-Darth: as in import and then use?

src/find/matchers/type_matcher.rs
Crypto-Darth: Done.
Had used it as a type before changing to hashset because of clippy warning.

src/find/matchers/type_matcher.rs
Crypto-Darth: Done.
Had used it as a type before changing to hashset because of clippy warning.

src/find/matchers/type_matcher.rs
Crypto-Darth: Done.

Crypto-Darth:
@tavianator let me know if it's good to go :)

Crypto-Darth:
@tavianator Please confirm if good to go.

src/find/matchers/type_matcher.rs
```diff
@@ -72,16 +87,46 @@ impl Matcher for XtypeMatcher {
...
+        self.file_type.contains(&file_type)
+    }
+}
+
+fn type_creator(type_string: &str, mode: &str) -> Result<HashSet<FileType>, Box<dyn Error>> {
+    let mut file_types = std::collections::HashSet::new();
+
+    if type_string.contains(',') {
+        for part in type_string.split(',') {
+            let trimmed = part.trim();
```
tavianator, line 109: Does GNU find trim these?

Crypto-Darth: image
Guess not.
Will remove it

Crypto-Darth: Done.

tavianator: Well that case is two separate arguments, you'd have to try `-type "f, d"`

Crypto-Darth: image

sylvestre:
thank you and sorry for the latency :(

Crypto-Darth:
> thank you and sorry for the latency :(

No worries all good I understand :)

Issue #523, opened 2025-04-04
fix blocks.div_ceil(len, bs)


sylvestre:
Might as well make the same fix right above: https://github.com/uutils/findutils/blob/5235d782916f51b31ba9e1f41fae717ba6dfd871/src/find/matchers/printf.rs#L396

Side note: I think the code here is wrong, but that can be fixed in a different PR.  It should be more like

```rust
let blocks = meta()?.blocks();
let len = blocks * STANDARD_BLOCK_SIZE;
let bs = if *large_blocks { 1024 } else { 512 };
blocks.div_ceil(len, bs)
```

_Originally posted by @tavianator in https://github.com/uutils/findutils/pull/522#discussion_r2027152154_

PR #529, opened 2025-04-06
Fix: Use div_ceil with proper blocks computation

KartikDua1504:
Fixes: #523

### DIFF
--- a/src/find/matchers/printf.rs
+++ b/src/find/matchers/printf.rs
@@ -390,23 +390,18 @@ fn format_directive<'entry>(
 
         FormatDirective::Blocks { large_blocks } => {
             #[cfg(unix)]
-            let blocks = meta()?.blocks();
+            let len = meta()?.blocks() * STANDARD_BLOCK_SIZE;
             #[cfg(not(unix))]
-            // Estimate using a ceiling division by the block size.
-            let blocks = (meta()?.len() + STANDARD_BLOCK_SIZE - 1) / STANDARD_BLOCK_SIZE;
+            let len = meta()?.len();
 
             // GNU find says it returns the number of 512-byte blocks for %b,
             // but in reality it just returns the number of blocks, *regardless
             // of their size on the filesystem*. That behavior is copied here,
             // even though it's arguably not 100% correct.
-            if *large_blocks {
-                // Ceiling divide in half.
-                blocks.div_ceil(2)
-            } else {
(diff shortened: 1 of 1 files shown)


KartikDua1504:
@sylvestre I have made the changes, but the intention of the code change were not clear to me and were ambiguous in the issue description. Could you please review and let me know if some other changes are required.

src/find/matchers/printf.rs
```diff
@@ -393,7 +393,13 @@ fn format_directive<'entry>(
             let blocks = meta()?.blocks();
             #[cfg(not(unix))]
             // Estimate using a ceiling division by the block size.
-            let blocks = (meta()?.len() + STANDARD_BLOCK_SIZE - 1) / STANDARD_BLOCK_SIZE;
+            let blocks = {
+                let blocks = meta()?.blocks();
+                let len = blocks * STANDARD_BLOCK_SIZE;
+                let bs = if *large_blocks { 1024 } else { 512 };
```
tavianator, line 399: This is now wrong as it's dividing the result again by 2 in the next statement.  I really meant to replace the whole thing with

```rust
FormatDirective::Blocks { large_blocks } => {
    #[cfg(unix)]
    let len = meta()?.blocks() * STANDARD_BLOCK_SIZE;
    #[cfg(not(unix))]
    let len = meta()?.len();

    let bs = if *large_blocks { 1024 } else { 512 };
    blocks.div_ceil(len, bs)
        .to_string()
        .into()
}
```

tavianator (changes requested):

tavianator (changes requested):
This is still wrong, you're only changing the `#[cfg(not(unix))]` statement.  You need to change the *whole* match guard

```rust
FormatDirective::Blocks { large_blocks } => {
    ...
}
```

src/find/matchers/printf.rs
KartikDua1504: Thanks for the suggestion, I have made the changes. It's my first time contributing to open-source so i made a mistake. Could you please check it now, and let me know if its correct?

tavianator (approved):

sylvestre:
Please run cargo fmt on it

KartikDua1504:
Sir, I formatted the document. Sorry for the inconvenience caused.

tavianator (approved):

PR #531, opened 2025-04-08
find: Fix -printf interactions with -L

tavianator:
- %l should only trigger for broken symlinks

- %Y should follow symlinks


sylvestre:
could you please add high level tests to make sure we don't regress ? 
thanks

tavianator:
There are bfs and probably GNU tests for this.  Unfortunately I probably won't have time soon to add uutils test cases

sylvestre:
Sure but i still would to have tests in this repo, much easier for development
thanks!

PR #532, opened 2025-04-08
find: Fix -fstype with stacked mounts

tavianator:
If you stack two mounts on top of each other, e.g.

    # mount -t ext4 /dev/sdX /mnt
    # mount -t f2fs /dev/sdY /mnt

then read_fs_list() will return two entries for that mount point.  The
later one is the one with the correct type, since the later mount is on
top of the earlier one.


Crypto-Darth:
image

Some minor clippy warnings here.

PR #533, opened 2025-04-09
find: Fix -execdir + working incorrectly for the root path

milkory:
This command should work now.

```
find / -maxdepth 0 -execdir echo {} +
```


sylvestre:
Could you please add a test to make sure we don't regress? Thanks

sylvestre (changes requested):
needs a test

milkory:
Yes, I'll add some later.

milkory:
I have added a test on `process_dir` and another test for `execdir +` in root directory.

PR #534, opened 2025-04-12
change "Error:" to "find:" + Testcase fix

Crypto-Darth:
Changed the error format of 
```Error: .......```
to
```find: .......```
to match GNU and also help pass testsuite.

Also fixed an error message which was failing test in GNU Testsuite and it is now passing (`tests/find/user-group-max.sh`)


Crypto-Darth:
Thank you :)

PR #536, opened 2025-04-14
implement locate and updatedb

### LINKED ISSUE #60: locate and updatedb
We should add implementations of both.

Qelxiros:
closes #60

Given that locate has largely been replaced by (m|p)locate, I was unable to find a version of locate packaged for my system (latest Arch Linux), and the packaged version of updatedb does not use the LOCATE02 db format. When running updatedb from source, I encountered errors that were not immediately fixable. As such, this implementation is largely untested, but it's better to have something here to be improved upon.


src/locate/main.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
```
sylvestre, line 1: please add the license header on all files :)

src/locate/mod.rs
```diff
@@ -0,0 +1,433 @@
+use std::{
```
sylvestre, line 1: don't hesitate to add some comments :)

sylvestre:
cool, thanks :)
a bunch of tasks are failing

sylvestre:
add please add tests :)

Qelxiros:
A couple notes:
- User flags are ignored for updatedb. GNU/BSD/Apple implementations all use a shell script, so they have direct access to su. I'm not sure if there's a better solution from rust, but I left it alone for now. 
- Performance is pretty significantly worse than GNU for updatedb. It might be a problem with find, since updatedb delegates most of the work to it. Locate is about on par with GNU for the one test that I did. I'd love another set of eyes to look for improvements.
- I'll fix coverage issues tomorrow.

src/db_tests.rs
```diff
@@ -0,0 +1,79 @@
+// Copyright 2017 Google Inc.
```
sylvestre, line 1:
```suggestion
```

src/db_tests.rs
```diff
@@ -0,0 +1,79 @@
...
+//
```
sylvestre, line 2: and tests should go in tests/ :)

Cargo.toml
```diff
@@ -11,14 +11,18 @@ authors = ["uutils developers"]
...
+clap = { version = "4.5", features = ["env"] }
 faccess = "0.2.4"
 walkdir = "2.5"
 regex = "1.11"
 onig = { version = "6.4", default-features = false }
 uucore = { version = "0.0.30", features = ["entries", "fs", "fsext", "mode"] }
 nix = { version = "0.30", features = ["fs", "user"] }
 argmax = "0.3.1"
+itertools = "0.14.0"
+quick-error = "2.0.1"
```
sylvestre, line 23: please switch to thiserror
quick-error hasn't been maintained for a while:
https://crates.io/crates/quick-error

src/locate/main.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright 2017 Google Inc.
```
sylvestre, line 1: please remove this line

src/locate/main.rs
```diff
@@ -0,0 +1,17 @@
...
+#[cfg(not(windows))]
+fn main() {
+    let args = std::env::args().collect::<Vec<String>>();
+    let strs: Vec<&str> = args.iter().map(std::convert::AsRef::as_ref).collect();
+    std::process::exit(findutils::locate::locate_main(strs.as_slice()));
+}
+
+#[cfg(windows)]
+fn main() {
+    println!("locate is unsupported on Windows");
```
sylvestre, line 16: maybe add why ? :)

Cargo.toml
```diff
@@ -11,14 +11,18 @@ authors = ["uutils developers"]
...
+uutests = "0.0.30"
```
sylvestre, line 24: do you use any of the uutests functions?

tests/db_tests.rs
```diff
@@ -0,0 +1,198 @@
+// Copyright 2017 Google Inc.
```
sylvestre, line 1: it would be nice to check the output of the commands and use uutests functions for that

Cargo.toml
Qelxiros: Will do. Worth noting that `uucore` explicitly recommends `quick-error` (https://github.com/uutils/coreutils/blob/main/src/uucore/src/lib/mods/error.rs#L156-L157) and several coreutils use it over `thiserror`. I'd be happy to open an issue/PR over there if you'd like to migrate to `thiserror` there as well.

sylvestre: Yes please :)
I was going to open good first bugs for this!

tests/db_tests.rs
Qelxiros: the uutests functions really want a multicall binary like coreutils, which isn't what we have here. I'd be interested in doing a larger refactor to bring findutils more in line with coreutils, but that probably shouldn't be this pr. should I just remove uutests as a dependency (to address your previous comment) and add uutests everywhere once this pr lands?

sylvestre: Yes please :)

src/updatedb/mod.rs
```diff
@@ -0,0 +1,335 @@
+// Copyright 2017 Google Inc.
```
sylvestre, line 1: Same, please remove that line and the others

Qelxiros: You asked me in https://github.com/uutils/findutils/pull/536#discussion_r2041546326 to add the license header. Should I just be removing the Google line? If so, the most recent commit should be good. If I'm misunderstanding, let me know.

Code7R:
Is there a reason for the kill switch which blocks operation on Windows?

Proper locate tool may be really interesting in the Windows world, where one would not have to install mingw or cygwin environments.

Qelxiros:
@Code7R The current implementation of `locate` relies on the UNIX-only `OsStr::from_bytes`. It might be possible to work around this on Windows with `OsString::from_wide`, but that at least requires a new allocation, and I'm not well-versed enough with Windows encodings to confidently identify how to properly convert from the `&[u8]` (read in from the db) to `&[u16]` (required by `from_wide`). As for `updatedb`, There's nothing explicitly preventing it from running on Windows, but I can't say with certainty that it doesn't make any UNIX-only assumptions.

Qelxiros:
@sylvestre any updates here?

sylvestre:
sorry, let me have a look
i just rebased your pr

PR #540, opened 2025-04-26
Implement `-context`

### LINKED ISSUE #375: Implement `-context`
refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Mode-Bits.html

```
Test: -context pattern
True if file’s SELinux context matches the pattern pattern. 
The pattern uses shell glob matching.

This predicate is supported only on find versions compiled with 
SELinux support and only when SELinux is enabled.
```

fu050409:
Closes #375

To avoid using C bindings, I wrote the detection logic for SELinux in Rust by referring to [libselinux](https://github.com/SELinuxProject/selinux/blob/main/libselinux/src/init.c), and used `xattr` crate to capture the SELinux extension for matching during the matching stage. `xattr` only supported on unix platforms so it was marked as `[target.'cfg(target_os = "linux")'.dependencies]`.

I'm still looking for a suitable way to write the test.


sylvestre:
looks like you missed my comment here: https://github.com/uutils/findutils/issues/375#issuecomment-2786658447

sylvestre:
https://github.com/uutils/coreutils/blob/main/src/uucore/src/lib/features/selinux.rs i have been working on this and to use it on findutils

also, there is that crate https://codeberg.org/koutheir/selinux
Sorry you wasted your time :(

fu050409:
> looks like you missed my comment here: [#375 (comment)](https://github.com/uutils/findutils/issues/375#issuecomment-2786658447) :(

@sylvestre I feel so sorry that GitHub doesn't seem to have pushed this comment to me correctly🥺, I didn't realize you had done some work on it and I didn't deliberately ignore your efforts.

fu050409:
> also, there is that crate https://codeberg.org/koutheir/selinux Sorry you wasted your time :(

I noticed the crate as well, but I personally don't think we should try to use more c bindings in findutils, it might be more appropriate to use more pure Rust code. Wdyt?

koutheir:
> I noticed the crate as well, but I personally don't think we should try to use more c bindings in findutils, it might be more appropriate to use more pure Rust code. What do you think?

**Short answer:** either (1) use the `selinux` crate, or (2) move this implementation-dependent logic into its own crate.

Looking at the code, I see multiple hard-coded paths, constants, and logic, all relative to the current implementation details of SELinux; details that could change anytime in subtle ways that could break this Rust implementation. It seems like bad design to me to have that non-trivial amount of logic inside this program. In my opinion, that logic belongs inside a dedicated crate that abstracts it away from this program. Now of course, I prefer that logic to be written in Rust too, but for the moment, such a crate doesn't exist, as far as I know. The closest thing to it is a Rust crate (`selinux`/`selinux-sys`) that calls into the official C implementation (`libselinux`) in order to perform all that logic. On one hand, this implies that using that crate will pull two more Rust crates and one more C shared library as dependencies, but on the other hand, we are not introducing a dependency that is not already in the original C implementation of this program.

*Full disclosure:* I'm the maintainer of the `selinux` and `selinux-sys` crates.

sylvestre:
a bunch of changes for this PR:

1. String::from_utf8(attr) keeps the trailing NUL - security.selinux is NUL-terminated, so patterns like *:s0 won't match. Strip it: attr.strip_suffix(b"\0").unwrap_or(&attr).
2. Gating on enforce is wrong - permissive mode (enforce==0) still has valid contexts. -context should work whenever SELinux is enabled, not enforcing.
3. new() returning Err aborts the whole command on non-SELinux/permissive hosts (e.g. -context x -o -name y). GNU is best-effort; don't hard-fail at parse time.
4. buf.parse::<i32>()? is fragile if /enforce has a trailing newline - use buf.trim().parse().
5. xattr::get follows symlinks unconditionally - ignores -H/-L/-P / WalkEntry follow mode. At least add a comment.
6. writeln!(...).unwrap() violates the no-unwrap/no-panic rule - use let _ = writeln!(…);.
7. Use nix's typed constant nix::sys::statfs::SELINUX_MAGIC instead of FsType(nix::libc::SELINUX_MAGIC) — avoids a fs_type_t cast mismatch across targets.
8. Tests missing (blocker) - add a parser test (-context w/ missing arg errors) and a runtime-gated integration test.

PR #546, opened 2025-05-07
ci: use `ubuntu-latest` instead of `ubuntu-20.04` for `release` workflow

cakebaker:
This PR uses `ubuntu-latest` instead of `ubuntu-20.04` in the `release` workflow. `ubuntu-20.04` has been removed from Github according to https://github.com/actions/runner-images/issues/11101 .

Please note that `release.yml` is usually auto-generated by `dist`. However, the changes of this PR are done manually as there is no newer version of `dist` available. The future of the project is currently unclear as the company behind it apparently run into financial problems.

Thanks to @Qelxiros who mentioned this issue in https://github.com/uutils/findutils/pull/536#issuecomment-2808407853 !

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,7 +47,7 @@ on:
 jobs:
   # Run 'dist plan' (or host) to determine what tasks we need to do
   plan:
-    runs-on: "ubuntu-20.04"
+    runs-on: "ubuntu-latest"
     outputs:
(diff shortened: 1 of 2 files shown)


cakebaker:
Ok, I resolved it with the workaround described in https://github.com/axodotdev/cargo-dist/issues/1760#issuecomment-2834549910

PR #553, opened 2025-06-01
matches/exec: create path integrity check before execution

### LINKED ISSUE #518: find: `-execdir` should ensure secure `PATH` variable
`-execdir` should check the `PATH` variable at parsing, and fails if it contains non-absolute directory names.

For example, this should fail:
```
export PATH=$PATH:.
find . -execdir echo hi ';'
```

refer: https://www.gnu.org/software/findutils/manual/html_node/find_html/Single-File.html

sami-daniel:
Fixes #518

Create function check_path_integrity which checks
the integrity of the path, with the following criteria:

- Only absolute paths
- Only non-empty paths
- Only directories in path

If the PATH env var does not pass the criteria,
an error is returned warning about the specific
path part that caused that error.

It was not possible to write tests for the changes, only to check if there were regressions. To test
the changes, it would be necessary to change the
PATH at test runtime. The only way to do this is
through the operating system API wrapper provided
in std::env::set_var. In order to use this API,
you must ensure that no other thread is reading
or writing from any place other than a single
place, thus avoiding data races and concurrency.
Unfortunately, the command engine that we currently use to execute other commands together with find
comes from std::process::Command, and it uses the
PATH to find the command to be executed. lazy_static would not work in this case either. A possible
solution would be to change the CIs to run only
`cargo test -- test-threads=1`, but I think the
tradeoff would not be worth it for such a simple
change. Besides the fact that other tests are
reading variables that we just changed, then
probably many executables would not be found


sylvestre:
could you also please add tests (with error mgmt)? thanks

sami-daniel:
> > could you also please add tests (with error mgmt)? thanks
> 
> done

Ok, I will fix these errors that are occurring on Windows before sending another push.

sami-daniel:
Since we cannot test in the exec.rs file, I couldn't test (directly) the `check_path_integrity` function. So, the codecov will probably point out an error at this point. However, tests have been added to validate the correct behavior of the `Matcher`s creation to work around this issue.

tests/exec_unit_tests.rs
```diff
@@ -23,6 +24,7 @@ use findutils::find::matchers::{Matcher, MatcherIO};
 mod common;
 
 #[test]
+#[serial(path_addicted)]
```
tavianator, line 27: "addicted"?

src/find/matchers/exec.rs
```diff
@@ -13,6 +14,27 @@ use std::process::Command;
...
 
+fn check_path_integrity() -> Result<(), Box<dyn Error>> {
+    let path_dirs = env::var("PATH")?;
+    for dir_entry in env::split_paths(&path_dirs) {
+        // We can securely unwrap (or expect) the value of dir_entry string
+        // conversion on message error cause the env::var returns an VarError
+        // variant that indicates if the variable (in this case PATH) contains
+        // invalid Unicode data.
+        let dir_entry_str = dir_entry.to_str().expect("Unexpected conversion error");
+        if !dir_entry.is_absolute() || dir_entry.is_file() || dir_entry_str.is_empty() {
```
tavianator, line 25:
```suggestion
        if !dir_entry.is_absolute() || dir_entry.as_os_str().is_empty() {
```

I don't think the `is_file()` check is necessary, GNU find doesn't do it.

tests/exec_unit_tests.rs
```diff
@@ -336,3 +347,334 @@ fn multi_set_exit_code_if_command_fails() {
     matcher.finished_dir(Path::new("test_data/simple"), &mut matcher_io);
     assert!(matcher_io.exit_code() == 1);
 }
+
+// -- tests for checking path integrity
+// SAFETY:
+// The only use of unsafe was for the std::env::set_var function.
+// Safety is guaranteed because all tests that directly depend
```
tavianator, line 354: Not really true, `set_var` is unsafe in a multi-threaded program unless you can guarantee that no other thread is accessing the environment at all (not just `PATH`).

For unit tests, I think we should just allow passing the `PATH` in.  For integration tests, we can spawn `find` with a new environment without calling `env::set_var()`.

src/find/matchers/exec.rs
sami-daniel: Ok

src/find/matchers/exec.rs
```diff
@@ -13,6 +14,27 @@ use std::process::Command;
 
 use super::{Matcher, MatcherIO, WalkEntry};
 
+fn check_path_integrity() -> Result<(), Box<dyn Error>> {
+    let path_dirs = env::var("PATH")?;
```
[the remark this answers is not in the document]
sami-daniel: As described in the [discussion](https://github.com/uutils/findutils/pull/553#discussion_r2137921828) where we see the extent of the `PATH` control, it is also determined that we cannot assume UTF-8 encoding for the variable (perhaps the system is not UTF-8). `env::var_os` is the best way. Just documenting.

src/find/matchers/exec.rs
```diff
@@ -248,6 +271,189 @@ impl Matcher for MultiExecMatcher {
...
-/// No tests here, because we need to call out to an external executable. See
-/// `tests/exec_unit_tests.rs` instead.
-mod tests {}
+/// Only tests related to path checking here, because we need to call out to an external executable. 
+/// See `tests/exec_unit_tests.rs` instead.
+mod check_path_tests {
+    use super::*;
+    use std::{fmt::Debug, os::unix::ffi::OsStringExt};
+
+    // unwrap_err requires T to implement Debug trait
```
tavianator, line 280: Might as well just `derive(Debug)` IMO

src/find/matchers/exec.rs
```diff
@@ -13,11 +14,32 @@ use std::process::Command;
...
+                return Err(format!(
+                    "The PATH environment variable contains non-absolute paths or empty paths. \
+                    Segment that caused the error: '{}'",
+                    dir_entry.as_os_str().to_string_lossy()
+                )
+                .into());
+            }
+        }
+    } else {
+        return Err("PATH environment variable is not defined.".into());
```
tavianator, line 30: This shouldn't be an error.  If `PATH` is unset, there is a default value to use, which in C is [`confstr(_CS_PATH)`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/confstr.html).  Not sure what it should be on Windows.

sami-daniel: I didn't find any wrapper or high level API around this. I guess we should use the libc crate for it

src/find/matchers/exec.rs
```diff
@@ -217,7 +266,417 @@ impl Matcher for MultiExecMatcher {
...
+        // Platform-specific non-UTF8 tests
+        #[cfg(unix)]
+        #[test]
+        #[ignore]
+        fn single_exec_matcher_does_not_reject_non_utf8_unix() {
+            let result = SingleExecMatcher::new_with_path(
+                "echo",
+                &["test"],
+                true,
+                Some(OsString::from_vec(vec![
```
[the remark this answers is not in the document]
sami-daniel: I think it works as a documentation about the expected behavior. Like, the engine should work even in environments that don't use utf8 as encoding. Let's suppose, on a machine where the PATH is stored as EBCDIC, the check should work normally. But since we can't test this directly...

src/find/matchers/exec.rs
sami-daniel: The problem is _CS_PATH may not be available on non-posix compliant system. I guess we should use `libc::confstr(...)` on Unix and on Windows we can use `C:\Windows\system32;C\Windows` as a `PATH` fallback.

sami-daniel: Haha, we don't need to use `libc::confstr` or do any further validation just remove the else branch

tavianator: Good point!  Generally the default `$PATH` is only absolute paths, nothing to worry about.  On Windows I think by default the current directory is searched.

PR #593, opened 2025-11-11
xargs: add support for -E option

brian-pane:


src/xargs/mod.rs
```diff
@@ -637,6 +640,38 @@ where
...
+                None => Ok(None),
+                Some(arg) => {
+                    if arg.arg == self.eof_delimiter {
+                        Ok(None)
+                    } else {
+                        Ok(Some(arg))
+                    }
+                }
+            },
+        }
```
cakebaker: I think you can simplify the code a bit by using `map`. Something like:
```suggestion
        self.reader.next().map(|opt| {
            opt.and_then(|arg| {
                if arg.arg == self.eof_delimiter {
                    None
                } else {
                    Some(arg)
                }
            })
        })
```

PR #603, opened 2025-11-26
Ignore -E/-e flag when a delimiter is specified

brian-pane:
(This fixes an incompatibility found by the Gnu findutils test suite)

### DIFF
--- a/src/xargs/mod.rs
+++ b/src/xargs/mod.rs
@@ -812,7 +812,7 @@ fn validate_positive_usize(s: &str) -> Result<usize, String> {
     }
 }
 
-fn normalize_options<'a>(options: &'a Options, matches: &'a clap::ArgMatches) -> Options {
+fn normalize_options(options: Options, matches: &clap::ArgMatches) -> Options {
     let (max_args, max_lines, replace) =
         match (options.max_args, options.max_lines, &options.replace) {
             // These 3 options are mutually exclusive.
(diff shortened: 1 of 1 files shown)


src/xargs/mod.rs
```diff
@@ -812,10 +812,17 @@ fn validate_positive_usize(s: &str) -> Result<usize, String> {
     }
 }
 
+#[allow(clippy::type_complexity)]
```
sylvestre: can we avoid this please? thanks

brian-pane: The alternatives that come to mind are:
1. Instead of returning a tuple from that function, return a new `Options` object.
2. Or pass in the `Options` object as a `mut` reference and modify it.

Would either of those be a good stylistic fit for the rest of the program?

sylvestre: 4 was probably already too much :)

sylvestre: it can be removed now, no ?

src/xargs/mod.rs
```diff
@@ -824,10 +821,10 @@ fn normalize_options<'a>(
                 // If `replace`, all matches in initial args should be replaced with extra args read from stdin.
                 // It is possible to have multiple matches and multiple extra args, and the Cartesian product is desired.
                 // To be specific, we process extra args one by one, and replace all matches with the same extra arg in each time.
-                (Some(1), None, &options.replace)
+                (Some(1), None, options.replace.clone())
```
sylvestre: is the clone really necessary ?

PR #620, opened 2026-01-28
find: implement {} path substitution for -exec

Erickkaranja:
A utility_name or argument containing only the two characters "{}" shall be replaced by the current pathname.

### DIFF
--- a/src/find/matchers/exec.rs
+++ b/src/find/matchers/exec.rs
@@ -53,8 +53,7 @@ impl SingleExecMatcher {
 
 impl Matcher for SingleExecMatcher {
     fn matches(&self, file_info: &WalkEntry, _: &mut MatcherIO) -> bool {
-        let mut command;
-        if &self.executable == "{}" {
+        let mut command = if &self.executable == "{}" {
             command = Command::new(file_info.path());
         } else {
             command = Command::new(&self.executable);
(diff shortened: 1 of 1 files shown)


sylvestre:
and needs a test and rebase

src/find/matchers/exec.rs
```diff
@@ -53,7 +53,11 @@ impl SingleExecMatcher {
 
 impl Matcher for SingleExecMatcher {
     fn matches(&self, file_info: &WalkEntry, _: &mut MatcherIO) -> bool {
-        let mut command = Command::new(&self.executable);
+        let mut command = if &self.executable == "{}" {
```
sylvestre: `self.executable == "{}"` should be enough, no need for the `&`

src/find/matchers/exec.rs
```diff
@@ -53,7 +53,11 @@ impl SingleExecMatcher {
...
+            Command::new(file_info.path())
+        } else {
+            Command::new(&self.executable)
```
sylvestre: if the exec fails, the error message will print `Failed to run {}` rather than the actual path
could you please fix that too?
and it needs a test

sylvestre:
sorry but it needs to be rebased

PR #643, opened 2026-03-19
Refactor compat test scripts and add Docker support for local testing

kevinburkesegment:
Extract shared logic (build_rust, check_total, print_summary,
generate_result_json) into util/common.sh. Make test directories
configurable via GNU_DIR/BFS_DIR env vars so scripts work in both
CI and local/Docker environments. Add Dockerfile.compat and
util/docker-compat.sh for running the full compat suites locally.

Simplify the CI workflow: remove empty "Extract testing info" step,
hoist CARGO_INCREMENTAL=0 to workflow-level env, drop ||: that masked
script errors, pass explicit GNU_DIR/BFS_DIR, and remove redundant
artifact download-and-compare steps.

### DIFF
--- a/.github/workflows/compat.yml
+++ b/.github/workflows/compat.yml
@@ -2,6 +2,9 @@ on: [push, pull_request]
 
 name: External-testsuites
 
+env:
+  CARGO_INCREMENTAL: 0
+
 jobs:
   gnu-tests:
     permissions:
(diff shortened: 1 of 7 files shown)


.github/workflows/compat.yml
```diff
@@ -59,54 +59,13 @@ jobs:
         with:
           name: gnu-result
           path: gnu-result.json
-      - name: Download artifacts (gnu-result and gnu-test-report)
```
kevinburkesegment, line 62: NOTE: The previous download-and-compare steps downloaded artifacts from
the *current* run (${{ github.run_id }}), then compared them against
the same logs already on disk -- producing "PASS +0 / FAIL +0" every
time (confirmed in upstream CI logs). To do actual regression detection,
this would need to download artifacts from the main branch's *latest*
run (e.g. via dawidd6/action-download-artifact). The artifact uploads
above are preserved for the upload-annotations job.

sylvestre:
sorry, it needs to be rebased

PR #647, opened 2026-03-27
fix: replace {} placeholder in -exec utility name

### LINKED ISSUE #614: `find -exec {} [args...] \;` fails
revision: 245623ecc6a19537b9f1fd94cb917653f170291d

```sh
cd /
find /usr/bin -name pwd -exec {} -P ';'  # /usr/bin/pwd -P
```

GNU and BSD show
```
/
```
and exit zero, whereas uutils fails:
```
Failed to run {}: No such file or directory (os error 2)
```

As per <https://pubs.opengroup.org/onlinepubs/9799919799.2024edition/utilities/find.html>, this is valid.

> <b>-exec&nbsp;</b><i>utility_name&nbsp;</i><b>[</b><i>argument</i>&nbsp;...<b>]&nbsp;;</b>
> …
> A <i>utility_name</i> or <i>argument</i> containing only the two characters <tt>"{}"</tt> shall be replaced by the current pathname.

mvanhorn:
## Summary

Fixed -exec to replace {} in the utility_name position, not just in arguments. Previously `find -exec {} \;` passed the literal string "{}" to Command::new.

## Why this matters

Per POSIX (`find -exec utility_name [argument ...] ;`), a utility_name containing only "{}" should be replaced with the current pathname. GNU find and BSD find both handle this correctly. uutils find failed with "No such file or directory" because {} was never substituted in the executable field.

Repro from #614:
```sh
find /usr/bin -name pwd -exec {} -P ';'
# Expected: runs /usr/bin/pwd -P
# Got: Failed to run {}: No such file or directory (os error 2)
```

## Changes

- `src/find/matchers/exec.rs`: Extracted `parse_arg` helper from the duplicated split logic. Changed `SingleExecMatcher.executable` from `String` to `Arg` so it undergoes the same {} replacement as arguments. Updated error formatting to use the resolved executable path.

## Testing

All 25 existing tests pass. Verified compilation with `cargo check` and `cargo clippy` (no warnings). Tested manually with `cargo run --bin find -- /usr/bin -name pwd -exec {} -P \;`.

Fixes #614

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/src/find/matchers/exec.rs
+++ b/src/find/matchers/exec.rs
@@ -18,8 +18,17 @@ enum Arg {
     LiteralArg(OsString),
 }
 
+fn parse_arg(s: &str) -> Arg {
+    let parts = s.split("{}").collect::<Vec<_>>();
+    if parts.len() == 1 {
+        Arg::LiteralArg(OsString::from(s))
+    } else {
+        Arg::FileArg(parts.iter().map(OsString::from).collect())
+    }
+}
+
 pub struct SingleExecMatcher {
-    executable: String,
+    executable: Arg,
     args: Vec<Arg>,
     exec_in_parent_dir: bool,
 }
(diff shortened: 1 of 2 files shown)


jmr:
What do you do with `foo{}bar` in the first position? According to POSIX, this is unspecified, but GNU will substitute there.

Also, probably a good idea to add some tests.  Claude didn't do that automatically?

sylvestre:
Indeed, it needs tests

mvanhorn:
Added a test in da847c6. On the `foo{}bar` question - the current implementation already handles embedded placeholders via the `FileArg(parts)` pattern (splits on `{}` and joins with the pathname), so `foo{}bar` in any position resolves to `foo/path/to/filebar`. This matches GNU find behavior.

sylvestre (changes requested):
sorry, it needs to be rebased :/

mvanhorn:
Appreciate the merge @sylvestre. Nice to have {} substitution in the utility_name position matching POSIX and GNU find now.

mvanhorn:
Appreciate you taking this, @sylvestre. The {} placeholder expanding in the -exec utility name matches GNU find behavior now.

PR #650, opened 2026-04-06
find: implement -ok and -okdir

### LINKED ISSUE #8: FeatureComplete: implement ok[dir] predicate

jmr:
-ok is like -exec ... ; but prompts the user on stderr before each invocation and only runs the command if the response is affirmative (starts with 'y' or 'Y').  -okdir is the corresponding -execdir variant. Only the ';' terminator is accepted: POSIX does not define -ok ... + and GNU find rejects it.

Response source: GNU find reads from /dev/tty so that the user's answer always comes from the real terminal even when stdin is occupied (e.g. `find -files0-from - -ok rm {} \;` reads paths from stdin, so responses cannot also come from there).  /dev/tty is the POSIX name for a process's controlling terminal and exists on all Unix-like systems.  We open /dev/tty only when stdin is itself a terminal (std::io::stdin().is_terminal() == true).  When stdin is a pipe or file, we read from stdin directly — matching BSD find and making scripted use (and integration tests via pipe_in()) work naturally without any special environment variable.  On Windows, or when /dev/tty cannot be opened, we likewise fall back to stdin.

Implementation: rather than a separate OkMatcher that duplicated SingleExecMatcher's fields, constructor, and exec logic, the interactive prompt is folded into SingleExecMatcher behind an `interactive: bool` field.  `SingleExecMatcher::new_interactive()` constructs the -ok/-okdir variant; `matches()` adds a guarded block that builds a GNU-find-style prompt ("< executable arg... >? ") and calls `matcher_io.confirm()` before executing.  If the user declines, the expression is false and the command is not run.  This keeps the arg-parsing, path resolution, current_dir logic, and error handling in one place so bug fixes apply to both -exec and -ok.

Dependencies::confirm() trait method: abstracts prompt+read so matchers remain testable without a real terminal; FakeDependencies uses a VecDeque of preset responses.

Tests:
- Unit tests (exec_unit_tests.rs): confirmed executes command, declined skips command and returns false, confirmed but command fails returns false, -okdir runs in parent dir
- Parser unit tests (matchers/mod.rs): missing-arg errors, missing semicolon, correct parse, confirm/decline via FakeDependencies
- Integration tests (test_find.rs): pipe_in() makes stdin a pipe so is_terminal() returns false and responses are read from the pipe; "y"/"Y"/"yes"/" y" run command, "n" and empty response skip command, -okdir runs in parent dir, prompt format verified, missing-semicolon error

Closes https://github.com/uutils/findutils/issues/8.

### DIFF
--- a/src/find/matchers/exec.rs
+++ b/src/find/matchers/exec.rs
@@ -22,6 +22,7 @@ pub struct SingleExecMatcher {
     executable: String,
     args: Vec<Arg>,
     exec_in_parent_dir: bool,
+    interactive: bool,
 }
 
 impl SingleExecMatcher {
@@ -29,6 +30,23 @@ impl SingleExecMatcher {
         executable: &str,
         args: &[&str],
         exec_in_parent_dir: bool,
+    ) -> Result<Self, Box<dyn Error>> {
+        Self::new_impl(executable, args, exec_in_parent_dir, false)
+    }
+
+    pub fn new_interactive(
+        executable: &str,
+        args: &[&str],
+        exec_in_parent_dir: bool,
+    ) -> Result<Self, Box<dyn Error>> {
+        Self::new_impl(executable, args, exec_in_parent_dir, true)
+    }
+
+    fn new_impl(
+        executable: &str,
+        args: &[&str],
+        exec_in_parent_dir: bool,
+        interactive: bool,
     ) -> Result<Self, Box<dyn Error>> {
         let transformed_args = args
             .iter()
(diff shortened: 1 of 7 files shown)


jmr:
The code missing in the failing coverage checks looks similar to existing code where there is no failure case possible.

https://github.com/uutils/findutils/blob/1266b7ffd1f4d056a7beef2068d2a0af52745d5c/src/find/matchers/exec.rs#L46

sylvestre:
well done :)

PR #656, opened 2026-04-09
fix: Handle write failures when printing --help/--version

### LINKED ISSUE #622: find/xargs --help/--version >/dev/full panics
```
$ target/debug/find --version > /dev/full

thread 'main' (19061) panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: No space left on device (os error 28)
```

Xylphy:
Fixes #622 

### Summary
Make `print_help` and `print_version` write via the `Dependencies` output and return `Result`; propagate errors from `do_find` so `find_main` exits with a non-zero code instead of panicking.


sylvestre (changes requested):
sorry, needs to be rebased

sylvestre:
please run cargo fmt

Xylphy:
@sylvestre Done 🙂

src/find/mod.rs
```diff
@@ -1601,4 +1606,48 @@ mod tests {
 
         assert_eq!(rc, 0);
     }
+    #[test]
```
sylvestre: i would prefer a test in tests/ instead

PR #668, opened 2026-04-28
xargs: accept hyphenated -I and -E values

kevinburke:
GNU xargs treats the arguments to -I, -E, and -e as values even when they begin with a hyphen. Clap was parsing those strings as options instead, so invocations like 'xargs -I -_' failed before xargs could process input.

Allow hyphen-leading values for those options and add regression coverage for replacement and EOF markers that start with '-'.


cakebaker:
Thanks for your PR!

PR #686, opened 2026-05-18
xargs: add --show-limits diagnostics

kevinburke:
Among other things, Debian maintainers use the output of xargs --show-limits to set --max-chars appropriately for subsequent commands.

Accept GNU-compatible --show-limits and print command-line size diagnostics before reading input or building a command. The output includes the Maximum length line parsed by Debian maintainer scripts.

The regression test links to Debian python3.13's
libPVER-minimal.prerm.in, where 3.13.5-2+deb13u2 parses xargs --show-limits, and to GNU findutils 4.10.0's xargs/xargs.c, where the diagnostic text is emitted.

Reuse the existing system command-size calculation for both the default limiter and the diagnostic output, and add an integration test for the parsed limit.

### DIFF
--- a/src/xargs/mod.rs
+++ b/src/xargs/mod.rs
@@ -33,6 +33,7 @@ mod options {
     pub const NULL: &str = "null";
     pub const REPLACE: &str = "replace";
     pub const REPLACE_I: &str = "replace-I";
+    pub const SHOW_LIMITS: &str = "show-limits";
     pub const VERBOSE: &str = "verbose";
 }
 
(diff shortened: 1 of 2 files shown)


sylvestre:
we have a difference here:
```

  echo "=== GNU ==="; printf 'a\nb\n' | /usr/bin/xargs --show-limits echo RAN 2>/dev/null
  echo "=== PR ===";  printf 'a\nb\n' | ./target/release/xargs --show-limits echo RAN 2>/dev/null
```

sylvestre (changes requested):
see the last comment

sylvestre:
@kevinburke ping?

PR #699, opened 2026-06-04
find: match GNU prompt format for -ok/-okdir

sylvestre:
The -ok/-okdir prompt rendered the full substituted command, but GNU find prints a fixed, abbreviated form: "< executable ... pathname > ? " (literal ..., a space before ?, and always the full entry path even for -okdir). Match it exactly and update the prompt-format test.

### DIFF
--- a/src/find/matchers/exec.rs
+++ b/src/find/matchers/exec.rs
@@ -83,20 +83,16 @@ impl Matcher for SingleExecMatcher {
         };
 
         if self.interactive {
-            let rendered_args: Vec<String> = self
-                .args
-                .iter()
-                .map(|arg| match arg {
-                    Arg::LiteralArg(a) => a.to_string_lossy().into_owned(),
-                    Arg::FileArg(parts) => parts
-                        .join(path_to_file.as_os_str())
-                        .to_string_lossy()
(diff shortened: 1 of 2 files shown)


src/find/matchers/exec.rs
```diff
@@ -83,20 +83,16 @@ impl Matcher for SingleExecMatcher {
...
-                })
-                .collect();
-            let mut prompt_parts = vec![self.executable.clone()];
-            prompt_parts.extend(rendered_args);
-            let prompt = format!("< {} >? ", prompt_parts.join(" "));
+            // GNU find prints a fixed, abbreviated prompt of the form
+            // "< executable ... pathname > ? ".  It does not render the
+            // substituted argument list, and always shows the full path of
+            // the entry being processed (even for -okdir, whose command runs
+            // with the "./basename" form).
```
hanbings, line 90: to my eyes, GNU's approach hides some of the detail, and I personally lean a little towards uutils's (even if its current implementation isn't all that polished) behavior of printing the full command. That said, I realize this might come at the cost of compatibility, so I'm honestly not sure

hanbings (approved):

PR #706, opened 2026-06-06
Implement Windows metadata for find -ls

caomengxuan666:
On Windows, ind -ls currently prints file attribute names such as rchive in the permission column and uses placeholder


sylvestre:
could you please rebase it? thanks

PR #708, opened 2026-06-06
xargs: clarify -x and -t help text

jmr:
The previous text implied -x only interacts with -L/-n and used imprecise phrasing. The new text mirrors the POSIX spec: terminate when a constructed command line would exceed the size limit rather than reducing the argument count.

Clarify that -t prints each command to stderr, rather than just "be verbose".


cakebaker:
Thanks!

PR #721, opened 2026-06-09
find: -ls: don't panic on an unmapped owner uid/gid

### LINKED ISSUE #717: find panics (unwrap None) on `-ls` for a file owned by an unmapped UID/GID
## Summary

Running `find <file> -ls` on a file whose owning UID (or GID) has no entry in `/etc/passwd` (or `/etc/group`) panics and exits 101. `-ls` resolves the owner name with `User::from_uid(...)`, which returns `Ok(None)` for an unmapped UID; the inner `.unwrap()` on that `Option` then panics. GNU `find` falls back to printing the numeric id instead. This is a routine condition: files from extracted archives that preserve foreign UIDs, NFS mounts, deleted accounts, and container images all produce unmapped owners.

## Steps to reproduce

```console
$ ./target/release/find /usr/local/lib/yottadb/r202/COPYING -ls
thread 'main' panicked at src/find/matchers/ls.rs:155:57:
called `Option::unwrap()` on a `None` value
$ echo $?
101
```

(The host file `/usr/local/lib/yottadb/r202/COPYING` is owned by UID 228, which has no `/etc/passwd` entry; any file with an unmapped owning UID reproduces it.)

## Expected behavior

Match GNU: print the entry using the numeric uid/gid, no crash, exit 0.

```console
$ /usr/bin/find /usr/local/lib/yottadb/r202/COPYING -ls
  6869904     36 -r--r--r--   1 228      root        34520 Dec  9  2024 /usr/local/lib/yottadb/r202/COPYING
$ echo $?
0
```

## Actual behavior

uutils `find` panics (`Option::unwrap()` on `None`) and exits 101. This triggers for any file whose owning UID lacks a `/etc/passwd` entry (line 155), and the identical sibling at line 159 triggers for any file whose GID lacks an `/etc/group` entry. The specific host path above is just one instance; the general portable condition (any unmapped owner/group) is what matters.

## Root cause

```rust
// src/find/matchers/ls.rs:153
let user = {
    let uid = metadata.uid();
    User::from_uid(Uid::from_raw(uid)).unwrap().unwrap().name
};
let group = {
    let gid = metadata.gid();
    Group::from_gid(Gid::from_raw(gid)).unwrap().unwrap().name
};
```

`User::from_uid` / `Group::from_gid` return `Ok(None)` when the UID/GID has no passwd/group entry. The outer `.unwrap()` handles the `Result`; the inner `.unwrap()` then panics on the `None` `Option`. 

_Found by source-audit of uutils/findutils._

leeewee:
Fixes #717

`find <file> -ls` panicked (exit 101) when the file's owning uid (or gid) had no `/etc/passwd` (`/etc/group`) entry: it resolved the name with `User::from_uid(...).unwrap().unwrap()`, and the inner unwrap aborts on the `Ok(None)` that an unmapped id returns. Unmapped owners are routine (extracted archives, NFS, deleted accounts, container images).

Resolve the uid/gid via `uucore::entries::uid2usr` / `gid2grp`, falling back to the numeric id when there is no entry, matching GNU find (which prints the number). Per review, this reuses the existing uucore helpers (which serialize the non-thread-safe `getpwuid`/`getgrgid` behind a mutex) rather than reimplementing the lookup.

Adds a `-ls` integration test that chowns a file to an unmapped id and checks the numeric fallback; it skips when the process lacks the privilege to do so.

### DIFF
--- a/src/find/matchers/ls.rs
+++ b/src/find/matchers/ls.rs
@@ -126,7 +126,6 @@ impl Ls {
         mut out: impl Write,
         print_error_message: bool,
     ) {
-        use nix::unistd::{Gid, Group, Uid, User};
         use std::os::unix::fs::{MetadataExt, PermissionsExt};
 
         let metadata = file_info.metadata().unwrap();
@@ -150,14 +149,13 @@ impl Ls {
         let permission =
             { format_permissions(metadata.permissions().mode() as uucore::libc::mode_t) };
         let hard_links = metadata.nlink();
-        let user = {
-            let uid = metadata.uid();
-            User::from_uid(Uid::from_raw(uid)).unwrap().unwrap().name
-        };
-        let group = {
-            let gid = metadata.gid();
-            Group::from_gid(Gid::from_raw(gid)).unwrap().unwrap().name
-        };
+        // Fall back to the numeric id when the uid/gid has no passwd/group entry
+        // (unmapped owner) — matching GNU find, which never crashes. uucore::entries
+        // serializes the non-thread-safe getpwuid/getgrgid behind a mutex.
+        let user =
+            uucore::entries::uid2usr(metadata.uid()).unwrap_or_else(|_| metadata.uid().to_string());
+        let group =
+            uucore::entries::gid2grp(metadata.gid()).unwrap_or_else(|_| metadata.gid().to_string());
         let size = metadata.size();
         let last_modified = {
             let system_time = metadata.modified().unwrap();
--- a/tests/test_find.rs
+++ b/tests/test_find.rs
@@ -1009,6 +1009,37 @@ fn find_ls() {
         .no_stderr();
 }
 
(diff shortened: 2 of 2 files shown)


src/find/matchers/ls.rs
```diff
@@ -11,6 +11,28 @@ use std::{
 
 use super::{Matcher, MatcherIO, WalkEntry};
 
+/// Resolve a uid to its login name, falling back to the numeric id when the uid
```
sylvestre: uucore already has these: `entries::uid2usr` / `gid2grp` (feature `entries`), which also serialize the non-thread-safe `getpwuid`/`getgrgid` behind a mutex. please reuse instead of reimplementing.

leeewee: Done — switched to `uucore::entries::uid2usr` / `gid2grp` with the numeric fallback and dropped the hand-rolled helpers. Thanks for the pointer!

PR #723, opened 2026-06-09
find: -printf: don't panic on a multibyte char after an octal escape

### LINKED ISSUE #720: find panics (str char-boundary) on a multibyte char after an octal escape in `-printf`
## Summary

`find -printf` panics on a format string where a `\` octal escape is immediately followed by a multibyte UTF-8 character (e.g. `'\0€\n'`). The escape parser peeks ahead 3 bytes for octal digits but checks only the byte length of the remaining string, not a char boundary, so it slices through the middle of the following multibyte char. This aborts with a Rust panic (exit code 101). GNU find accepts the same input, emitting the escaped byte then the literal character.

## Steps to reproduce

```console
$ mkdir -p /tmp/ftest
$ ./target/release/find /tmp/ftest -printf '\0€\n'
thread 'main' panicked at src/find/matchers/printf.rs:146:24:
byte index 3 is not a char boundary; it is inside '€' (bytes 1..4) of `0€`
$ echo $?
101
```

Other inputs that trigger the same panic: `'\1😀'`, `'\00é'`.

## Expected behavior

Match GNU: process the octal escape and print the following character literally, exiting 0.

```console
$ /usr/bin/find /tmp/ftest -printf '\0€\n'
$ echo $?
0
```

## Root cause

`parse_escape_sequence` calls `peek(OCTAL_LEN)` (3 bytes) when an octal digit follows `\`, and `peek` slices the string after checking only its byte length:

```rust
// src/find/matchers/printf.rs:141
fn peek(&self, count: usize) -> Result<&str, Box<dyn Error>> {
    if self.string.len() < count {
        return Err("Unexpected EOF".into());
    }

    Ok(&self.string[0..count]) // line 146: panics if `count` is not a char boundary
}
```

The guard `self.string.len() < count` only ensures there are enough *bytes*; it does not ensure that byte index `count` falls on a UTF-8 char boundary. When a multibyte character occupies bytes that span index 3 (e.g. `€` at bytes 1..4 of `0€`), `&self.string[0..3]` slices inside the char and panics.

_Found by source-audit of uutils/findutils._

leeewee:
Fixes #720

`find -printf '\NNN'` octal escapes were parsed by slicing a fixed 3 bytes off the format string. When fewer than 3 octal digits are followed by a multibyte character (e.g. `-printf '\0€'`), that 3-byte slice lands inside the multibyte char and panics on a UTF-8 char-boundary assertion (exit 101); GNU `find` exits 0.

The octal escape is now parsed from its leading octal digits only (1 to 3, all ASCII) and the parser advances by their byte length, so a following multibyte char is never sliced into.

This also fixes `\1`..`\7`: previously only `\0` was recognised and `\1`..`\7` fell through to the single-character escape table and errored, instead of being treated as octal — now matching GNU `find`.

Verified byte-for-byte against GNU for `\0€`, `\1😀`, `\00é`, `\101`, `\012X`. Added a regression unit test.

### DIFF
--- a/src/find/matchers/printf.rs
+++ b/src/find/matchers/printf.rs
@@ -167,16 +167,21 @@ impl FormatStringParser<'_> {
         // Try parsing an octal sequence first.
         let first = self.front()?;
         if first.is_digit(OCTAL_RADIX) {
-            if let Ok(code) = self.peek(OCTAL_LEN).and_then(|octal| {
-                u32::from_str_radix(octal, OCTAL_RADIX).map_err(std::convert::Into::into)
-            }) {
-                // safe to unwrap: .peek() already succeeded above.
-                let octal = self.advance_by(OCTAL_LEN).unwrap();
-                return match char::from_u32(code) {
-                    Some(c) => Ok(FormatComponent::Literal(c.to_string())),
-                    None => Err(format!("Invalid character value: \\{octal}").into()),
-                };
-            }
+            // A GNU octal escape is 1 to 3 octal digits. Consume only the leading
+            // octal digits (which are ASCII), rather than slicing a fixed 3 bytes
+            // that can land inside a following multibyte char.
+            let octal: String = self
+                .string
+                .chars()
+                .take(OCTAL_LEN)
+                .take_while(|c| c.is_digit(OCTAL_RADIX))
+                .collect();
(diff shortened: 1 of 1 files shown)


sylvestre:
thanks for also proposing PR !
could you please also add a test in :
tests/test_find.rs

PR #731, opened 2026-06-12
find: -printf: advance past a full char in advance_one to avoid a multibyte panic

### LINKED ISSUE #730: find: -printf: advance_one still panics on a multibyte char (after %, \, or a time directive)
#723 fixed the `-printf` octal-escape `peek` path for multibyte characters, but `advance_one` still byte-slices and panics when a multibyte character follows a `%` conversion, a `\` escape, or a `%A`/`%C`/`%T` time-field intro.

## Steps to reproduce

```console
$ mkdir d; touch d/a
$ find d -printf '%€\n'
thread 'main' panicked at src/find/matchers/printf.rs:152:35:
byte index 1 is not a char boundary; it is inside '€' (bytes 0..3) of `€`
$ echo $?
101
Same panic via find d -printf '\€\n' and find d -printf '%A€' (and %C, %T).
```

## Expected (GNU)

```
$ /usr/bin/find d -printf '%€\n'
%€
$ echo $?
0
```

## Cause

After reading a full char, `advance_one` drops it by slicing one byte:

```rust
// src/find/matchers/printf.rs:150
fn advance_one(&mut self) -> Result<char, Box<dyn Error>> {
    let c = self.front()?;
    self.string = &self.string[1..];   // line 152: byte 1 is mid-char for a multibyte `c`
    Ok(c)
}
```

When `c` is multibyte, byte index 1 isn't a char boundary, so the slice panics.

leeewee:
Fixes #730

`advance_one` read a full `char` via `front()` but then dropped it by slicing one byte (`&self.string[1..]`). When that char is multibyte (e.g. `€` after a `%` conversion, a `\` escape, or a `%A`/`%C`/`%T` time directive), byte index 1 is not a char boundary, so the slice panicked:

```console
$ mkdir d; touch d/a
$ find d -printf '%€\n'
thread 'main' panicked at src/find/matchers/printf.rs:152:35:
byte index 1 is not a char boundary; it is inside '€' (bytes 0..3) of `€`
$ echo $?
101
```

(also `\€` and `%A€` / `%C€` / `%T€`)

This advances by the leading char's UTF-8 length (`c.len_utf8()`) instead of a fixed `1`, so a multibyte char routes through the normal handling instead of aborting — the same multibyte-aware approach #723 applied to `peek`. Unrecognized escapes / invalid time specifiers keep their existing behavior (a multibyte char now behaves exactly like its ASCII equivalent, e.g. `\€` like `\q`).

Added a regression test (`test_parse_multibyte_char_after_directive`).

### DIFF
--- a/src/find/matchers/printf.rs
+++ b/src/find/matchers/printf.rs
@@ -152,7 +152,7 @@ impl FormatStringParser<'_> {
 
     fn advance_one(&mut self) -> Result<char, Box<dyn Error>> {
         let c = self.front()?;
-        self.string = &self.string[1..];
+        self.string = &self.string[c.len_utf8()..];
         Ok(c)
     }
 
@@ -725,6 +725,16 @@ mod tests {
                 FormatComponent::Literal("é".to_owned()),
             ]
         );
+    }
+
+    #[test]
+    fn test_parse_multibyte_char_after_directive() {
+        assert_eq!(
+            FormatString::parse("%€").unwrap().components,
+            vec![FormatComponent::Literal("€".to_owned())]
+        );
+        assert!(FormatString::parse("\\€").is_err());
+        assert!(FormatString::parse("%A€").is_err());
     }
 
     #[test]
--- a/tests/test_find.rs
+++ b/tests/test_find.rs
@@ -588,6 +588,22 @@ fn find_printf_width_too_large() {
         .stderr_contains("find: Invalid format width");
 }
 
+#[test]
+fn find_printf_multibyte_char_after_directive() {
+    ucmd()
+        .args(&["./test_data/simple", "-maxdepth", "0", "-printf", "%€\\n"])
+        .succeeds()
+        .stdout_only("€\n");
+    ucmd()
+        .args(&["./test_data/simple", "-maxdepth", "0", "-printf", "\\€\\n"])
+        .fails()
+        .stderr_contains("find: Invalid escape sequence");
+    ucmd()
+        .args(&["./test_data/simple", "-maxdepth", "0", "-printf", "%A€"])
+        .fails()
+        .stderr_contains("find: Invalid time specifier");
+}
+
 #[cfg(unix)]
 #[test]
 fn find_perm() {


sylvestre:
On windows:

```

failures:

---- find_printf stdout ----
bin: "D:\\a\\findutils\\findutils\\target\\debug\\find.exe"
run: D:\a\findutils\findutils\target\debug\find.exe .\test_data\simple -sorted -printf %f %d %h %H %p %P %y

bin: "D:\\a\\findutils\\findutils\\target\\debug\\find.exe"
run: D:\a\findutils\findutils\target\debug\find.exe a -printf %A+

thread 'find_printf' (8780) panicked at tests\test_find.rs:486:5:
Output did not match expected timestamp format



```

leeewee:
> On windows:
> 
> ```
> 
> failures:
> 
> ---- find_printf stdout ----
> bin: "D:\\a\\findutils\\findutils\\target\\debug\\find.exe"
> run: D:\a\findutils\findutils\target\debug\find.exe .\test_data\simple -sorted -printf %f %d %h %H %p %P %y
> 
> bin: "D:\\a\\findutils\\findutils\\target\\debug\\find.exe"
> run: D:\a\findutils\findutils\target\debug\find.exe a -printf %A+
> 
> thread 'find_printf' (8780) panicked at tests\test_find.rs:486:5:
> Output did not match expected timestamp format
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

The `find_printf` failure is unrelated to this PR — the change here is only in `advance_one`, a no-op for ASCII, so it can't affect `%A+`. But digging in, it's a real pre-existing bug: `%A+/%C+/%T+` use chrono `%.f,` which drops the fraction when it isn't full-precision, so e.g. a whole-second timestamp prints `…:200` instead of GNU's `…:20.0000000000`. That's what makes the test flaky on Windows (coarse file times).

leeewee:
I opened #757 to fix the underlying issue — the %A+ fraction wasn't fixed-width, which is what made this test flaky on Windows.

sylvestre:
"This branch cannot be rebased due to conflicts"
could you please rebase it? thanks

leeewee:
Rebased onto the latest main.

PR #747, opened 2026-06-20
release: pin macOS/Windows targets to live runners

sylvestre:
Releases are published with **no binary assets** (e.g. [`0.9.0`](https://github.com/uutils/findutils/releases/tag/0.9.0), `0.9.1`), unlike [uutils/diffutils](https://github.com/uutils/diffutils/releases/tag/v0.5.0) which attaches a full set of platform binaries.

Root cause: the cargo-dist 0.28 `plan` step requests runner labels that **GitHub has retired**:


cakebaker:
Hm, wouldn't it make sense to update `cargo-dist` instead? The current version is `0.32.0`.

sylvestre:
Yes right sorry

PR #762, opened 2026-07-15
find: report an unparsable -newerXt year instead of panicking

### LINKED ISSUE #759: find: `-newerXt` with a non-ASCII-digit year aborts instead of erroring
The `-newerXt` time options (`-newermt`, `-newerat`, `-newerct`, `-newerBt`) parse the year field of their date argument with `.parse::<i32>().unwrap()` (`src/find/matchers/mod.rs:384`). The year is captured by a regex whose `\d` class matches **Unicode** decimal digits (the `regex` crate defaults to Unicode mode, so `\d` == `\p{Nd}`), but `str::parse::<i32>()` accepts only ASCII digits. So a year written with four non-ASCII decimal digits (Arabic-Indic, Devanagari, fullwidth, …) passes the regex and then panics in `parse`, aborting the process with exit 101.

```console
$ find . -newermt 'jan 01, ٠٠٠٠'
thread 'main' panicked at src/find/matchers/mod.rs:384:56:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }
$ echo $?
101
```

Any four non-ASCII decimal digits in the year field trigger it — the digit system is irrelevant:

```console
$ find . -newermt 'jan 01, ０００１'   # fullwidth digits
thread 'main' panicked at src/find/matchers/mod.rs:384:56: ...
$ find . -newermt 'jan 01, ००१२'      # Devanagari digits
thread 'main' panicked at src/find/matchers/mod.rs:384:56: ...
```

GNU findutils `find` rejects the same argument cleanly:

```console
$ /usr/bin/find . -newermt 'jan 01, ٠٠٠٠'
/usr/bin/find: I cannot figure out how to interpret ‘jan 01, ٠٠٠٠’ as a date or time
$ echo $?
1
```

## Root cause

`parse_date_str_to_timestamps` (`src/find/matchers/mod.rs`):

```rust
let regex_pattern =
    r"^(?P<month_day>\w{3} \d{2})?(?:, (?P<year>\d{4}))?(?: (?P<time>\d{2}:\d{2}:\d{2}))?$";
let re = Regex::new(regex_pattern);
if let Some(captures) = re.ok()?.captures(date_str) {
    ...
    let year = captures
        .get(2)
        .map_or(now.year(), |m| m.as_str().parse().unwrap());   // <-- mod.rs:384: panics
    ...
    let datetime = NaiveDateTime::parse_from_str(&date_time_str, "%b %d, %Y %H:%M:%S").ok()?;
    ...
}
```

`\d{4}` matches four Unicode decimal digits, but `m.as_str().parse::<i32>()` (the type is inferred from the `now.year(): i32` branch) rejects non-ASCII digits with `ParseIntError { kind: InvalidDigit }`, which the `.unwrap()` turns into a panic.

_Found by our static analysis tooling._

leeewee:
Fixes #759.

`-newerXt` (`-newermt`/`-newerat`/`-newerct`/`-newerBt`) parsed the year field with `.parse().unwrap()`. The capturing regex's `\d{4}` matches Unicode decimal digits (the `regex` crate defaults to Unicode mode), but `str::parse` accepts only ASCII, so a year of four non-ASCII digits (e.g. `find . -newermt 'jan 01, ٠٠٠٠'`) passed the regex and then panicked, aborting with exit 101.

This folds a failed year parse into the function's existing `None` path, so the argument is reported as `find: I cannot figure out how to interpret ‘…’ as a date or time` (exit 1) — matching GNU findutils and what uutils already does for other malformed dates. Added a regression assertion to `parse_date_str_to_timestamps_test`.

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -379,9 +379,10 @@ fn parse_date_str_to_timestamps(date_str: &str) -> Option<i64> {
                 m.as_str().to_string()
             });
         // If no year input.
-        let year = captures
-            .get(2)
-            .map_or(now.year(), |m| m.as_str().parse().unwrap());
+        let year = match captures.get(2) {
+            Some(m) => m.as_str().parse().ok()?,
+            None => now.year(),
+        };
         // If the user does not enter a specific time, it will be filled with 0
         let time_str = captures.get(3).map_or("00:00:00", |m| m.as_str());
         let date_time_str = format!("{month_day}, {year} {time_str}");
@@ -1840,6 +1841,9 @@ mod tests {
             .and_utc()
             .timestamp_millis();
         assert_eq!(none_date_timestamps, Some(now_but_zero_hour_min_sec));
+
+        // A year of non-ASCII decimal digits must be rejected, not panic.
+        assert_eq!(parse_date_str_to_timestamps("jan 01, ٠٠٠٠"), None);
     }
 
     #[test]


cakebaker:
Thanks!

PR #768, opened 2026-07-16
locate: ignore an unrepresentable --max-database-age instead of panicking

### LINKED ISSUE #766: locate: a large `--max-database-age` aborts via `TimeDelta::days` overflow
`locate` compares the database age against `--max-database-age` (default 8) with `TimeDelta::days(config.max_age as i64)` (`src/locate/mod.rs:622`). `chrono`'s `TimeDelta::days` **panics** when the resulting duration overflows its internal range, so a large numeric argument (which `clap` accepts as a `usize`) aborts the process inside `chrono` instead of being rejected or clamped.

```console
$ printf '\0LOCATE02\0\0/foo\0' > /tmp/db.bin
$ locate -d /tmp/db.bin --max-database-age 1000000000000 foo
thread 'main' panicked at .../chrono-0.4.45/src/lib.rs:717:17:
overflow in TimeDelta::days
$ echo $?
101
```

## Root cause

`src/locate/mod.rs` (the age-check inside the DB reader loop):

```rust
let delta = now - modified;
if delta
    > TimeDelta::days(config.max_age as i64)   // <-- mod.rs:622: panics on overflow
{
    eprintln!("{}: warning: database ‘{}’ is more than {} days old ...", ...);
}
```

`config.max_age: usize` comes straight from `--max-database-age`. `chrono`'s `TimeDelta::days(d)` panics when `d` days can't be represented.

leeewee:
Fixes #766.

`locate` compared the database age against `--max-database-age` using `TimeDelta::days(config.max_age as i64)`. `chrono`'s `TimeDelta::days` is the panicking constructor, so a large but syntactically valid value aborted the process:

```console
$ printf '\0LOCATE02\0\0/foo\0' > /tmp/db.bin
$ locate -d /tmp/db.bin --max-database-age 1000000000000 foo
thread 'main' panicked at chrono-0.4.45/src/lib.rs:717:17:
overflow in TimeDelta::days
$ echo $?
101
```

The `as i64` cast was also wrong for values above `i64::MAX`: `usize::MAX` wrapped to `-1`, so a freshly built database was reported as too old:

```console
$ locate -d /tmp/db.bin --max-database-age 18446744073709551615 foo
locate: warning: database ‘/tmp/db.bin’ is more than 18446744073709551615 days old (actual age is 0.0 days)
```

This guards both conversions and treats an age that `i64`/`chrono` cannot represent as never exceeded, which matches what GNU `locate` effectively does: `set_max_db_age` only errors on `ERANGE`/invalid/trailing junk, and otherwise assigns the value into an `unsigned int warn_number_units`, so a huge age truncates and simply never warns.

The check moves into an `is_db_too_old` helper so it can be unit tested. The default (8) and ordinary values are unchanged — `--max-database-age 0` still warns.

### DIFF
--- a/src/locate/mod.rs
+++ b/src/locate/mod.rs
@@ -591,6 +591,15 @@ fn match_entry(entry: &CStr, config: &Config, patterns: &Patterns) -> bool {
     patterns_match && existence_matches
 }
 
+/// Whether a database of the given `age` is older than `max_age` days.
+/// A `max_age` too large for `i64`/`chrono` to represent can never be exceeded.
+fn is_db_too_old(age: TimeDelta, max_age: usize) -> bool {
+    match i64::try_from(max_age).ok().and_then(TimeDelta::try_days) {
+        Some(limit) => age > limit,
+        None => false,
+    }
+}
+
 fn do_locate(args: &[&str]) -> LocateResult<()> {
     let matches = uu_app().try_get_matches_from(args);
     match matches {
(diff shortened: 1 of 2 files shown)


sylvestre:
please add a test in tests/db_tests.rs for this

leeewee:
> please add a test in tests/db_tests.rs for this

Added, thanks!

PR #773, opened 2026-07-16
fix: don't panic in locate on a failed stdout write

### LINKED ISSUE #769: locate: panics when its output is redirected to /dev/full
Every one of `locate`'s stdout output paths uses `print!`/`println!`, which panic on a failed write.

The affected output paths (`src/locate/mod.rs`):

- **`--version`** — `print!("{}", app.render_version())` (`mod.rs:599`)
- **match output** — `println!("{}", …)` / `print!("{}\0", …)` (`mod.rs:649`/`651`)
- **`-c` / count** — `println!("{count}")` (`mod.rs:671`)
- **`-S` / `--statistics`** — the whole `Statistics::print` / `print_header`
  block, ~10 `println!` calls (`mod.rs:117`–`144`)

```console
$ locate --version > /dev/full
thread 'main' panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: No space left on device (os error 28)
$ echo $?
101

$ printf '\0LOCATE02\0\0/foo\0' > /tmp/db.bin
$ locate --statistics -d /tmp/db.bin foo > /dev/full
thread 'main' panicked at library/std/src/io/stdio.rs:1165:9: ...
$ locate -d /tmp/db.bin foo > /dev/full          # match output
thread 'main' panicked at library/std/src/io/stdio.rs:1165:9: ...
$ locate -c -d /tmp/db.bin foo > /dev/full        # count output
thread 'main' panicked at library/std/src/io/stdio.rs:1165:9: ...
$ echo $?
101
```

mvanhorn:
## Summary

Every one of `locate`'s stdout output paths used `print!`/`println!`, which panic on a failed write. Redirecting output to `/dev/full` (or having a downstream reader close the pipe) therefore aborts `locate` with a panic and exit code 101 instead of failing gracefully.

Closes #769.

## Why this matters

`locate` is a batch tool whose output is routinely piped (`locate foo | head`, `locate foo > file`). A write failure — a full disk, or the common `| head` case where the reader closes early — is an ordinary I/O condition, not a bug in the program. GNU `locate` reports the error and exits non-zero; panicking (exit 101, with a backtrace) is both wrong behavior and, under `python -O`-style optimized builds where `debug_assert`-adjacent asserts are stripped, would let the handler keep running on a dead stream.

The affected paths (all in `src/locate/mod.rs`): `--version`, the match output (`println!`/`print!("{}\0")`), `-c`/count, and the `-S`/`--statistics` block.

## Fix

Route every stdout write through `writeln!`/`write!` and propagate the `io::Error` via the existing `Error::IoErr(#[from] io::Error)` variant, so `locate_main` returns `code() == 1` instead of aborting. The two `Statistics` methods now return `LocateResult<()>`. The match-output path runs inside an iterator `inspect` closure (which returns `()`), so its first write error is captured in a local and propagated with `?` after iteration rather than restructuring the pipeline.

## Testing

`/dev/full` only exists on Linux, so I verified with a broken pipe (which Rust surfaces as `BrokenPipe`, since it ignores `SIGPIPE`):

- Built `locate` and a 30k-entry LOCATE02 DB. On `master`, `locate -d bigdb foo | true` panics (`thread 'main' panicked at .../io/stdio.rs:1166`, exit 101). With this change it exits 1.
- Normal behavior is unchanged: match output (`/foo/bar`, `/foo/baz`), `-c` count (`2`), `--version`, and `--statistics` all produce identical output.
- `cargo build`, `cargo fmt -- --check`, and `cargo clippy` are clean for the changed file (no new warnings).

I did not add an automated test: exercising a real write failure needs either an injectable writer (a broader refactor of the stdout call sites) or a somewhat fragile subprocess broken-pipe test. Happy to add one in whatever form fits the project's test conventions.

### DIFF
--- a/src/locate/mod.rs
+++ b/src/locate/mod.rs
@@ -113,35 +113,45 @@ impl Statistics {
         }
     }
 
-    fn print_header(&self, dbreader: &DbReader) {
-        println!(
+    fn print_header(&self, dbreader: &DbReader) -> LocateResult<()> {
+        writeln!(
+            io::stdout(),
             "Database {} is in the {} format.",
             dbreader.path.to_string_lossy(),
             dbreader.format,
-        );
+        )?;
+        Ok(())
     }
 
-    fn print(&self, dbreader: &DbReader) {
+    fn print(&self, dbreader: &DbReader) -> LocateResult<()> {
+        let mut stdout = io::stdout().lock();
         if let Ok(metadata) = fs::metadata(&dbreader.path) {
             if let Ok(time) = metadata.modified() {
                 let time: DateTime<Local> = time.into();
-                println!("Database was last modified at {}", time);
+                writeln!(stdout, "Database was last modified at {}", time)?;
             }
-            println!("Locate database size: {} bytes", metadata.len());
+            writeln!(stdout, "Locate database size: {} bytes", metadata.len())?;
         }
-        println!("Matching Filenames: {}", self.matches);
-        println!(
+        writeln!(stdout, "Matching Filenames: {}", self.matches)?;
+        writeln!(
+            stdout,
             "File names have a cumulative length of {} bytes",
             self.total_length
-        );
-        println!("Of those file names,\n");
-        println!("        {} contain whitespace,", self.whitespace);
-        println!("        {} contain newline characters,", self.newlines);
-        println!(
+        )?;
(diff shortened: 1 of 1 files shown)


sylvestre:
sorry, i didn't notice (and this PR had a clippy issue)

PR #784, opened 2026-07-17
find: reject invalid -newerXY predicates like GNU

### LINKED ISSUE #782: bug(find): Invalid predicates accepted
Hi, findutils mainteners

we confirmed the uu find, accepts invalid predicate flags, as we confirmed

```css
relunsec@relunsec:~/software/findutils/target/debug$ sudo ./find -neweraBcmty .
relunsec@relunsec:~/software/findutils/target/debug$ sudo /bin/find -neweraBcmty .
/bin/find: unknown predicate `-neweraBcmty'
```

l46983284-cpu:
## Summary

`find` accepted invalid predicates such as `-neweraBcmty` because the `-newerXY` parser matched a valid prefix instead of the whole argument. GNU find reports those as unknown predicates.

This change:

- anchors `-newerXY` matching so only exact `X∈{a,B,c,m}` / `Y∈{a,B,c,m,t}` forms parse
- reports unknown predicates with GNU-style wording: `unknown predicate \`…'` (as in the issue sample)
- adds unit + integration coverage for the issue #782 cases
- fixes a pre-existing unit test that incorrectly glued `-follow -newerXY` into one argv token (that only worked while the regex was unanchored)

Note: open PR #724 only rewords the unknown-predicate message for ordinary unrecognized args. This PR also fixes the real #782 root cause (prefix match on `-newerXY`) and uses the GNU backtick/apostrophe quote shape from the issue.

## Test plan

- [x] `cargo test --lib` (227 passed)
- [x] `cargo test --test test_find invalid_newerxy_predicate_is_rejected`
- [x] Integration harness covers `find -neweraBcmty .` failing with `unknown predicate \`-neweraBcmty'`

Fixes #782

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -399,6 +399,10 @@ fn parse_date_str_to_timestamps(date_str: &str) -> Option<i64> {
 /// X and Y are constrained to a/B/c/m and t.
 /// such as: "-neweraB" -> Some(a, B) "-neweraD" -> None
 ///
+/// The whole argument must match. A trailing suffix after a valid
+/// -newerXY form (for example -neweraBcmty) is not accepted; GNU find
+/// reports those as unknown predicates.
+///
 /// Additionally, there is support for the -anewer and -cnewer short arguments. as follows:
 /// 1. -anewer is equivalent to -neweram
 /// 2. -cnewer is equivalent to -newercm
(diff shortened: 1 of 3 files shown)


HackingRepo:
one job fails because of intermittent issues https://github.com/uutils/findutils/actions/runs/29574082323/job/87867034162

PR #788, opened 2026-07-21
xargs: Account for quoting on Windows for length limits

lhecker:
As reported at https://github.com/microsoft/coreutils/issues/160.

### DIFF
--- a/src/xargs/mod.rs
+++ b/src/xargs/mod.rs
@@ -141,8 +141,41 @@ impl Clone for LimiterCollection {
 #[cfg(windows)]
 fn count_osstr_chars_for_exec(s: &OsStr) -> usize {
     use std::os::windows::ffi::OsStrExt;
+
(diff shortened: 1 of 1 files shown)


src/xargs/mod.rs
```diff
@@ -141,8 +141,41 @@ impl Clone for LimiterCollection {
...
+                // Account for n+1 additional backslashes before a quote.
+                len += backslashes + 1;
+                backslashes = 0;
+            }
+            SPACE | TAB => {
+                quote = true;
+                backslashes = 0;
+            }
+            _ => backslashes = 0,
+        }
```
lhecker, line 170: Apologies if this code is quite fugly. I'd also wish Windows had a sane argv system. 🙁

lhecker:
What does that failing test mean?

sylvestre:
just that one more test is executed :)
probably added upstream?

PR #798, opened 2026-07-28
find: reject invalid UTF-8 in --files0-from

### LINKED ISSUE #781: bug(find): not error out when the --files0-from is invalid UTF8 like \xFF
Hi, findutils mainteners

i confirmed that the uu find do not output any errors when a.txt contains \xFF

```css
relunsec@relunsec:~/software/findutils/target/debug$ sudo ./find -files0-from a.txt
relunsec@relunsec:~/software/findutils/target/debug$ echo $'\xFF\nfile2.txt' > a.txt
relunsec@relunsec:~/software/findutils/target/debug$ sudo ./find -files0-from a.txt
relunsec@relunsec:~/software/findutils/target/debug$ sudo /bin/find -files0-from a.txt
/bin/find: ‘\377\nfile2.txt\n’: Invalid or incomplete multibyte or wide character
```

wtcpython:
Fixes #781.

`--files0-from` decoded each NUL-separated starting point as UTF-8 and silently discarded segments that were not valid UTF-8. As a result, non-UTF-8 input could produce no traversal and a successful exit status.

Propagate UTF-8 decoding errors instead of silently ignoring invalid starting points.

Add a regression test using the issue's non-UTF-8 input and verify that it fails without producing output.

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -1041,9 +1041,8 @@ fn parse_files0_args(config: &mut Config) -> Result<(), Box<dyn Error>> {
 
     let mut string_segments: Vec<String> = buffer_split
         .iter()
-        .filter_map(|s| std::str::from_utf8(s).ok())
-        .map(std::string::ToString::to_string)
-        .collect();
+        .map(|segment| std::str::from_utf8(segment).map(std::string::ToString::to_string))
+        .collect::<Result<_, _>>()?;
     // empty starting point checker
     if string_segments.iter().any(std::string::String::is_empty) {
         eprintln!("find: invalid zero-length file name");
--- a/tests/test_find.rs
+++ b/tests/test_find.rs
@@ -195,6 +195,20 @@ fn files0_file_basic_success() {
 }
 
 #[test]
+fn files0_invalid_utf8_fails() {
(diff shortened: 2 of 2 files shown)


src/find/matchers/mod.rs
```diff
@@ -1039,22 +1039,40 @@ fn parse_files0_args(config: &mut Config) -> Result<(), Box<dyn Error>> {
...
+    use std::os::unix::ffi::OsStringExt;
+
+    PathBuf::from(OsString::from_vec(bytes.to_vec()))
+}
+
+#[cfg(not(unix))]
+fn path_from_bytes(bytes: &[u8]) -> Result<PathBuf, Box<dyn Error>> {
+    Ok(PathBuf::from(std::str::from_utf8(bytes)?))
+}
+
```
sylvestre: i don't think it is an interesting test

wtcpython: Sorry — I over-scoped the fix by adding native-path handling beyond what the issue requires. I've simplified it to propagate the UTF-8 decoding error, removing the PathBuf refactor and both platform-specific helpers. The PR is now limited to the parser change and its regression test. Thanks for pointing this out.

sylvestre: i am not a fan of answering to an LLM in a comment... esp when it adds 0 value :(

PR #808, opened 2026-08-04
fix: add rust-version to Cargo.toml

xtqqczze:
Also update workflows to run on multiple os and rust-version.

verified with `cargo hack check --workspace --all-targets --feature-powerset --rust-version`


sylvestre:
still draft ?

xtqqczze:
@sylvestre ready

PR #817, opened 2026-08-07
find: don't panic on invalid-UTF-8 arguments

### LINKED ISSUE #816: `-printf` panics on an invalid-UTF-8 format string (exit 101)
A `-printf` format string that is not valid UTF-8 panics during argument collection, before the format parser is ever reached.

```console
$ find d -printf $'%\xff|\n'
thread 'main' panicked at ...: called `Result::unwrap()` on an `Err` value
$ echo $?
101
```

The panic is in `std::env::args`, which panics on non-UTF-8 arguments. Collecting via `std::env::args_os` and handling the invalid-UTF-8 case explicitly would let this produce a normal error (or match GNU, which accepts arbitrary bytes in the format string and passes them through).

MsfPablo:
## Summary

`find` used `std::env::args()`, which panics if any argument is not
valid UTF-8:

```console
$ find d -printf $'%\xff|\n'
thread 'main' panicked at ...: called `Result::unwrap()` on an `Err` value
$ echo $?
101
```

The panic happens during argument collection, before the `-printf`
format parser is ever reached.

This switches to `std::env::args_os()` and reports invalid UTF-8 as a
normal error (exit 1) instead of panicking:

```console
$ find d -printf $'%\xff|\n'
find: invalid UTF-8 was found in one of the arguments: %�|
$ echo $?
1
```

Full binary/byte passthrough (matching GNU find's behavior of
accepting arbitrary bytes) would need `find_main` and the whole
argument-parsing pipeline to switch from `&str` to `OsStr`, which is
a much larger refactor — happy to look into that separately if
maintainers want it, but this fixes the immediate panic/exit-101 bug
with a minimal change.

## Test plan

- [x] Added `find_printf_invalid_utf8_format_does_not_panic` integration test
- [x] `cargo test` — 226 + 60 + ... all passing (full suite green)
- [x] `cargo fmt --check` clean
- [x] Manually reproduced the original panic and confirmed the fix

Fixes #816

### DIFF
--- a/src/find/main.rs
+++ b/src/find/main.rs
@@ -11,14 +11,18 @@ fn main() {
     uucore::panic::mute_sigpipe_panic();
 
     let args = std::env::args_os()
-        .map(|arg| {
-            arg.into_string().unwrap_or_else(|invalid| {
+        .map(|arg| match arg.into_string() {
+            Ok(s) => s,
+            // GNU find treats an invalid-UTF-8 argument as a warning and
+            // continues with a lossy conversion (exiting 0), rather than
+            // aborting — so do the same instead of hard-erroring with exit 1.
+            Err(invalid) => {
                 eprintln!(
                     "find: invalid UTF-8 was found in one of the arguments: {}",
                     invalid.to_string_lossy()
                 );
-                std::process::exit(1);
-            })
+                invalid.to_string_lossy().into_owned()
+            }
         })
         .collect::<Vec<String>>();
     let strs: Vec<&str> = args.iter().map(std::convert::AsRef::as_ref).collect();
--- a/src/find/main.rs
+++ b/src/find/main.rs
@@ -10,7 +10,17 @@ fn main() {
     // the downstream software of the standard output stream closes the pipe and triggers a panic.
     uucore::panic::mute_sigpipe_panic();
 
-    let args = std::env::args().collect::<Vec<String>>();
+    let args = std::env::args_os()
+        .map(|arg| {
+            arg.into_string().unwrap_or_else(|invalid| {
+                eprintln!(
+                    "find: invalid UTF-8 was found in one of the arguments: {}",
+                    invalid.to_string_lossy()
+                );
+                std::process::exit(1);
+            })
+        })
+        .collect::<Vec<String>>();
     let strs: Vec<&str> = args.iter().map(std::convert::AsRef::as_ref).collect();
     let deps = findutils::find::StandardDependencies::new();
     std::process::exit(findutils::find::find_main(&strs, &deps));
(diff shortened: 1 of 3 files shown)


src/find/main.rs
```diff
@@ -10,7 +10,17 @@ fn main() {
...
 
-    let args = std::env::args().collect::<Vec<String>>();
+    let args = std::env::args_os()
+        .map(|arg| {
+            arg.into_string().unwrap_or_else(|invalid| {
+                eprintln!(
+                    "find: invalid UTF-8 was found in one of the arguments: {}",
+                    invalid.to_string_lossy()
+                );
+                std::process::exit(1);
```
Devel08: that is a warning in GNU findutils, thus exits with 0

MsfPablo:
You're right — thanks. GNU find treats this as a warning and exits 0, not 1. Fixed in 4dabff7: `main` now lossy-converts the arg and continues (preserving the warning), instead of `std::process::exit(1)`, so the exit code reflects only real errors like file-not-found.

Lossy-converting the `-printf` format then exposed a separate, pre-existing panic in the printf parser: `FormatStringParser::advance_one` byte-sliced `self.string[1..]` to drop the character it had just read, which panics whenever that character is multibyte — a lossy replacement char from invalid UTF-8, but also any non-ASCII char following a `%`-directive (e.g. `find -printf '%€'`). Fixed by slicing `self.string[c.len_utf8()..]` so one character is dropped rather than one byte.

Verified locally: `cargo build`, `cargo test --test test_find` (60 passed), `cargo clippy` clean. The test now expects success (exit 0) plus the warning, matching GNU.

src/find/matchers/printf.rs
```diff
@@ -152,7 +152,11 @@ impl FormatStringParser<'_> {
 
     fn advance_one(&mut self) -> Result<char, Box<dyn Error>> {
         let c = self.front()?;
-        self.string = &self.string[1..];
+        // Slice off one *character*, not one byte: byte slicing `[1..]` panics
```
sylvestre, line 155: please make this comment shorter

src/find/main.rs
```diff
@@ -10,7 +10,21 @@ fn main() {
     // the downstream software of the standard output stream closes the pipe and triggers a panic.
     uucore::panic::mute_sigpipe_panic();
 
-    let args = std::env::args().collect::<Vec<String>>();
+    let args = std::env::args_os()
+        .map(|arg| match arg.into_string() {
+            Ok(s) => s,
+            // GNU find treats an invalid-UTF-8 argument as a warning and
```
sylvestre, line 16: same, one line comment please

sylvestre:

MsfPablo:
Closing as superseded: commit e1927ac (#816 "find: avoid panic on a non-UTF-8 argument"), now on main, already resolves the underlying crash by collecting args via `args_os` + `OsString::into_string` and reporting the first invalid one as an error instead of panicking in `std::env::args`. This PR's panic fix is therefore redundant, and its remaining behavioral proposal (lossy-convert + continue, exit 0) is the opposite exit semantics from the approach #816 adopted (reject with an error, non-zero exit), so I won't push a competing design. Thanks for the report that motivated #816.

PR #818, opened 2026-08-07
-printf: warn on unrecognized directives/escapes instead of dropping or erroring

### LINKED ISSUE #815: -printf: unrecognized %X / \X directives should warn and print literally, not be dropped or  error
GNU findutils treats an unrecognized `-printf` directive as a warning and prints the directive text verbatim, exiting 0. uutils/findutils diverges in two ways.

### `%X` - the `%` is silently dropped

```console
$ # GNU findutils 4.10.0
$ find d -printf '%€|\n'
%€|
$ echo $?
0

$ # uutils
$ find d -printf '%€|\n'
€|
$ echo $?
0
```

The `%` disappears. Same for the ASCII case, e.g. `%q`.

### `\X` - uutils errors instead of warning

```console
$ # GNU: warns, prints `\€`, exit 0
$ # uutils: hard error, exit 1
```

MsfPablo:
## Summary

GNU findutils treats an unrecognized `-printf` directive/escape as a
warning and prints it literally, exiting 0. This diverged in two ways:

**`%X` — the `%` was silently dropped**
```console
$ # GNU findutils 4.10.0
$ find d -printf '%€|\n'
%€|

$ # uutils, before this PR
$ find d -printf '%€|\n'
€|
```

**`\X` — uutils errored instead of warning**
```console
$ # before this PR: hard error, exit 1
$ # after: warning + literal output, exit 0
```

Both cases now print a `find: warning: ...` message to stderr and emit
the directive/escape literally (with its `%`/`\` prefix), matching GNU.

Along the way, fixed `advance_one()`: it sliced the remaining format
string at a fixed 1-byte offset, which panics on multibyte characters
(e.g. `€`, as in the issue's own repro) immediately after `%` or `\`.
Switched to `char::len_utf8()`.

## Test plan

- [x] Added integration tests for both the multibyte (`%€`, `\€`) and
      the plain-ASCII unrecognized-directive case
- [x] Updated the two existing unit tests (`test_parse_escapes`,
      `test_parse_formatting`) that asserted the old (incorrect) behavior
- [x] `cargo test` — full suite green (226 unit + 61 integration tests)
- [x] `cargo fmt --check` clean
- [x] Manually verified output matches the GNU findutils transcript in the issue

Fixes #815

### DIFF
--- a/src/find/matchers/printf.rs
+++ b/src/find/matchers/printf.rs
@@ -152,7 +152,7 @@ impl FormatStringParser<'_> {
 
     fn advance_one(&mut self) -> Result<char, Box<dyn Error>> {
         let c = self.front()?;
-        self.string = &self.string[1..];
+        self.string = &self.string[c.len_utf8()..];
         Ok(c)
     }
 
@@ -203,7 +203,10 @@ impl FormatStringParser<'_> {
                 'v' => "\x0B",
                 '0' => "\0",
                 '\\' => "\\",
-                c => return Err(format!("Invalid escape sequence: \\{c}").into()),
+                c => {
+                    eprintln!("find: warning: unrecognized escape '\\{c}'");
+                    return Ok(FormatComponent::Literal(format!("\\{c}")));
+                }
             };
 
             Ok(FormatComponent::Literal(c.to_string()))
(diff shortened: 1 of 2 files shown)


sylvestre:
nice, thanks

PR #820, opened 2026-08-08
xargs: cap single-argument growth to avoid OOM on unterminated input

### LINKED ISSUE #770: bug(xargs): when -a is /dev/full, it do not error out instead allocates memory indefinetively
Hi, findutils mainteners

we want let you known, when passing to -a, /dev/full, it just hangs and tries allocate massive memory, until killed, we confirmed using free -mh

```css
relunsec@relunsec:~/software/findutils/target/debug$ sudo ./xargs -a /dev/full ls
^C
```

until killed by the linux kernel OOM or pressing CTRL+C, while the gnu correctly error out and abort not tries to allocate indefnitevly memory, unlike the uu find

```css
relunsec@relunsec:~/software/findutils/target/debug$ sudo /bin/xargs -a /dev/full ls
/bin/xargs: WARNING: a NUL character occurred in the input.  It cannot be passed through in the argument list.  Did you mean to use the --null option?
/bin/xargs: argument line too long
```

MsfPablo:
Closes #770.

## Problem

`xargs` reading from an input that never produces a delimiter — most notably `/dev/full`, which yields an endless stream of `NUL` bytes — accumulates a single argument without bound until the process is OOM-killed. Nothing bounded argument growth between delimiters, so a terminator-less (or non-matching-delimiter) stream could grow the reader buffer indefinitely.

A minimal repro (Linux, with `/dev/full`):

```
$ xargs -a /dev/full echo
# ... accumulates memory until OOM-killed
```

## Fix

Bound each accumulated argument by a multiple of the effective command-line character budget — the same budget the `-s` / system `ARG_MAX`-derived limiters already enforce — and report `argument line too long` once an argument exceeds it. This mirrors the error GNU `xargs` produces in this situation.

- New `MaxCharsCommandSizeLimiter::effective_max_chars(options_max_chars, env)` returns the per-command-line budget (min of user `-s` and the system limit, or the 128 KiB default), used both to construct the existing limiter and to derive the argument cap.
- `ARG_SIZE_OVERFLOW_MULTIPLIER = 4`: the cap is generously sized so that no argument the size limiters would ever accept can be rejected by the reader — it only ever trips on truly unbounded growth.
- Both `WhitespaceDelimitedArgumentReader` and `ByteDelimitedArgumentReader` gain a `max_arg_size` bound, checked as bytes are accumulated.
- The reader error type is widened from `io::Result` to a small `XargsError` enum so the `argument line too long` condition is distinguishable from an ordinary I/O error (e.g. `Interrupted`, `BrokenPipe`).

The byte-delimited reader previously relied on `BufReader::read_until` to loop internally; driving the buffered reader by hand (so the cap can be checked) surfaces `Interrupted` reads to the caller, so those are retried explicitly to preserve the previous behaviour.

## Verification

- `cargo build` — clean
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo fmt --check` — clean
- `cargo test` — 360 tests pass, 0 failures

Added two unit tests covering the cap directly (portable, no `/dev/full` dependency) so the behaviour is locked in regardless of platform:

- `test_whitespace_reader_caps_unbounded_argument` — a long run of non-whitespace with no terminator returns `ArgumentTooLarge` instead of looping forever.
- `test_byte_reader_caps_unbounded_argument` — a stream lacking the delimiter byte returns `ArgumentTooLarge` instead of looping forever.

Existing reader tests were updated to the new constructor signatures and the `XargsError` return type; the `Interrupted`-retried and `BrokenPipe`-propagation behaviours are still covered by `test_byte_delimited_reader` and `test_eof_argument_reader`.

`/dev/full` itself is Linux-only, so I kept the end-to-end repro on Linux rather than gating a test behind `cfg(target_os = "linux")`; the unit tests above exercise the same code path portably.

### DIFF
--- a/src/xargs/mod.rs
+++ b/src/xargs/mod.rs
@@ -1605,8 +1605,10 @@ mod tests {
             &b"aaa \nbbb\n"[..],
             &b"aaa \nbbb \n \t \n"[..],
         ] {
-            let mut reader =
-                WhitespaceDelimitedArgumentReader::new(ChunkReader::new(vec![Chunk::Data(input)]));
+            let mut reader = WhitespaceDelimitedArgumentReader::new(
+                ChunkReader::new(vec![Chunk::Data(input)]),
+                usize::MAX,
+            );
             assert_eq!(reader.next().unwrap().unwrap().arg, "aaa", "{input:?}");
             assert_eq!(reader.next().unwrap().unwrap().arg, "bbb", "{input:?}");
             assert_eq!(reader.next().unwrap(), None, "{input:?}");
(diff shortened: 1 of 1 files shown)


sylvestre:
please add a test in test_find

sylvestre:
jobs are failing

PR #822, opened 2026-08-08
xargs: strip `(os error N)` from --arg-file open failure

MsfPablo:
## Summary

When `--arg-file` points at a missing path, `xargs` used to surface the raw `io::Error` display, which appends ` (os error N)` for OS errors:

```
$ xargs --arg-file sdf echo
Error: Failed to open sdf: No such file or directory (os error 2)
```

GNU `xargs` omits that detail. This reuses `uucore::error::strip_errno` — the same helper `updatedb` already uses for the identical "strip the trailing `(os error N)` so the message reads cleanly" goal — so the output becomes:

```
Error: Failed to open sdf: No such file or directory
```

This is a one-line change at the arg-file open site plus the existing helper; no new abstraction.

Closes #811.

## Test plan

- [x] Added `xargs_arg_file_missing_strips_errno` integration test asserting: exit code 1, stderr contains `Failed to open`, contains `No such file or directory`, and does **not** contain `(os error`.
- [x] `cargo test` — full workspace suite green (find, locate, updatedb, xargs: 28/28 + doctests).
- [x] `cargo fmt --check` clean.
- [x] `cargo clippy --bin xargs` clean.
- [x] Manually reproduced the issue transcript and confirmed the `(os error N)` suffix is gone.

## Context

I went with `strip_errno` rather than a custom regex because `updatedb/mod.rs` already established it as the codebase pattern for exactly this (`strip the trailing "(os error N)" ...`). Happy to change the wording (e.g. GNU-style `xargs: sdf: No such file or directory`) if maintainers prefer a closer GNU message match — kept it minimal here since #811 only asks to drop the `(os error N)`.

This contribution was developed with AI assistance (Claude). I have reviewed and understand every line of the diff.

### DIFF
--- a/src/xargs/mod.rs
+++ b/src/xargs/mod.rs
@@ -15,6 +15,7 @@ use std::{
 };
 
 use clap::{crate_version, error::ErrorKind, Arg, ArgAction};
+use uucore::error::strip_errno;
 
 mod options {
     pub const COMMAND: &str = "COMMAND";
@@ -1184,7 +1185,10 @@ fn do_xargs(args: &[&str]) -> Result<CommandResult, XargsError> {
     builder_options.close_stdin = options.arg_file.is_none();
 
     let args_file: Box<dyn Read> = if let Some(path) = &options.arg_file {
-        Box::new(fs::File::open(path).map_err(|e| format!("Failed to open {path}: {e}"))?)
+        Box::new(
+            fs::File::open(path)
+                .map_err(|e| format!("Failed to open {path}: {}", strip_errno(&e)))?,
+        )
     } else {
         Box::new(io::stdin())
     };
--- a/tests/test_xargs.rs
+++ b/tests/test_xargs.rs
@@ -559,13 +559,17 @@ fn xargs_eof_with_delimiter() {
 
 #[test]
 fn xargs_arg_file_missing_strips_errno() {
-    // A missing --arg-file should fail with a clean "No such file or
-    // directory" message and *not* expose the trailing "(os error N)"
-    // detail — matching GNU xargs. See issue #811.
+    // A missing --arg-file should fail with a message naming the path and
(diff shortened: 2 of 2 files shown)


sylvestre:
Fails on windows:

```


---- xargs_arg_file_missing_strips_errno stdout ----
bin: "D:\\a\\findutils\\findutils\\target\\debug\\find.exe"
run: D:\a\findutils\findutils\target\debug\xargs.exe --arg-file /no/such/findutils-test-path echo

thread 'xargs_arg_file_missing_strips_errno' (7324) panicked at tests\test_xargs.rs:569:10:
'Error: Failed to open /no/such/findutils-test-path: The system cannot find the path specified.
' does not contain 'No such file or directory'


```

MsfPablo:
Thanks @sylvestre — good catch. Pushed a fix (b8fd422): the test hardcoded the Unix errno wording `"No such file or directory"`, which fails on Windows where the OS reports `"The system cannot find the path specified."`. Since the contract this test guards is only that the trailing `(os error N)` suffix is stripped (the OS wording is out of our control), it now asserts the portable parts — our `Failed to open <path>:` wrapper names the path and no `(os error` suffix leaks. Local `cargo test --test test_xargs` is green (28/28), `cargo clippy --bin xargs -D warnings` and `cargo fmt --check` clean.

Separately, the compat (BFS) job flags `gnu/files0_from_ok` as a new failure, but that one is pre-existing on `main` independent of this PR — the latest `main` External-testsuites run ([30980036584](https://github.com/uutils/findutils/actions/runs/30980036584), 2026-08-05, before this PR) also reports `[FAIL] gnu/files0_from_ok` in its BFS log. The "new" label comes from diffing against an older baseline artifact; the only change here is a one-line xargs `--arg-file` error-message string, which cannot touch `find -files0-from`.

sylvestre:
@MsfPablo please answering through your llm

PR #828, opened 2026-08-11
find: share one file handle per output path (fixes #439)

### LINKED ISSUE #439: Specifying the same file for output multiple times in find will overwrite the file by default.

MsfPablo:
Fixes #439.

## Problem

GNU `find` de-duplicates the files opened by its output predicates, so
`find . -fprint foo -fprint foo` writes each path to `foo` twice. uutils
opened a separate `File` for every occurrence of `-fprint`/`-fprintf`/
`-fprint0`/`-fls`, so each handle had its own file offset and the writes
overwrote one another — only one line per entry survived.

## Fix

`get_or_create_file` now caches opened output files on `Config`, keyed by the
path as written on the command line (matching GNU's literal-filename
de-duplication), and hands out a shared `Rc<File>`. Repeated references to the
same path therefore share a single file descriptor and a single file offset.
`Printer`, `Printf` and `Ls` take `Option<Rc<File>>` instead of `Option<File>`.
The first reference still truncates, as before.

## Tests

Two unit tests in `src/find/matchers/mod.rs`:

- `get_or_create_file_reuses_handle_for_same_path` — the same path yields the
  same `Rc`, and writes through both handles append rather than overwrite.
- `two_fprints_to_the_same_file_do_not_overwrite_each_other` — end-to-end
  through `build_top_level_matcher` with `-fprint out -fprint out`, asserting
  the entry appears twice.

Verified the regression test fails without the fix (reverting just the cache
lookup gives `left: "…abbbc\n…abbbc\n"` / `right: "…abbbc\n"`).

```
$ cargo test
test result: ok. 228 passed; 0 failed  (lib)
... all other suites ok, 0 failed
$ cargo clippy --all-targets --all-features -- -D warnings
    Finished `dev` profile
$ cargo fmt --all --check
(clean)
```

## AI disclosure

This change was written with the assistance of Claude (Anthropic). I have
reviewed the diff, and ran the build, the full test suite, clippy and rustfmt
locally as shown above.

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -65,7 +65,6 @@ use ls::Ls;
 use std::{
     error::Error,
     fs::{File, Metadata},
-    io::Read,
     path::{Path, PathBuf},
     rc::Rc,
     str::FromStr,
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -65,7 +65,9 @@ use ls::Ls;
 use std::{
     error::Error,
     fs::{File, Metadata},
+    io::Read,
     path::{Path, PathBuf},
+    rc::Rc,
     str::FromStr,
     time::SystemTime,
 };
@@ -440,10 +442,21 @@ fn parse_str_to_newer_args(input: &str) -> Option<(String, String)> {
     }
 }
 
-/// Creates a file if it doesn't exist.
-/// If it does exist, it will be overwritten.
-fn get_or_create_file(path: &str) -> Result<File, Box<dyn Error>> {
-    let file = File::create(path)?;
+/// Returns the output file for `path`, creating (and truncating) it the first
+/// time it is requested.
+///
+/// Later requests for the same path reuse the handle opened earlier, so that
+/// several output predicates writing to one file share a single file offset
+/// instead of overwriting each other.
+fn get_or_create_file(config: &mut Config, path: &str) -> Result<Rc<File>, Box<dyn Error>> {
+    if let Some(file) = config.output_files.get(path) {
+        return Ok(Rc::clone(file));
+    }
+
+    let file = Rc::new(File::create(path)?);
+    config
+        .output_files
+        .insert(path.to_string(), Rc::clone(&file));
     Ok(file)
 }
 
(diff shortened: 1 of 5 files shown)


sylvestre:
many jobs are failing

sylvestre:

src/find/matchers/mod.rs
```diff
@@ -440,10 +441,21 @@ fn parse_str_to_newer_args(input: &str) -> Option<(String, String)> {
...
-/// If it does exist, it will be overwritten.
-fn get_or_create_file(path: &str) -> Result<File, Box<dyn Error>> {
-    let file = File::create(path)?;
+/// Returns the output file for `path`, creating (and truncating) it the first
+/// time it is requested.
+///
+/// Later requests for the same path reuse the handle opened earlier, so that
+/// several output predicates writing to one file share a single file offset
+/// instead of overwriting each other.
+fn get_or_create_file(config: &mut Config, path: &str) -> Result<Rc<File>, Box<dyn Error>> {
```
sylvestre: keyed on the literal string, so `-fprint foo -fprint ./foo` still overwrites, no?
could you please check what GNU find does for that (and for a symlink)? if it dedups those too, maybe compare dev/ino instead

src/find/matchers/mod.rs
```diff
@@ -1908,25 +1922,66 @@ mod tests {
...
+
+        // Writes through both handles share one file offset, so neither
+        // overwrites the other.
+        writeln!(&*first, "one").unwrap();
+        writeln!(&*second, "two").unwrap();
+        assert_eq!("one\ntwo\n", fs::read_to_string(&path).unwrap());
+    }
+
+    #[test]
+    fn two_fprints_to_the_same_file_do_not_overwrite_each_other() {
```
sylvestre: please add a test in tests/test_find.rs too (with `-fprint` and `-fls`/`-fprintf` mixed on the same file)

src/find/mod.rs
```diff
@@ -32,6 +34,11 @@ pub struct Config {
     /// Whether the expression uses -ok or -okdir, which prompt on stderr and
     /// read the answer from stdin when there is no terminal.
     interactive_exec: bool,
+    /// Files opened by output predicates (-fprint, -fprintf, -fprint0, -fls),
+    /// keyed by the path given on the command line. Reusing one handle per
+    /// path means specifying the same output file more than once appends
+    /// rather than overwriting (see issue #439).
+    output_files: HashMap<String, Rc<File>>,
```
sylvestre: same, please make it shorter, no need for the issue number

PR #831, opened 2026-08-12
find: stream -files0-from input instead of reading it all into memory

### LINKED ISSUE #779: perf(find): when using -files0-from /dev/urandom or /dev/full, it consumes massive large memory
Hi, findutils mainteners

we confirmed when using

```css
relunsec@relunsec:~/software/findutils/target/debug$ ./find -files0-from /dev/urandom
```

will exhaust all your device memory instantly and output no output all, as i confirmed with `watch free -mh`

```css
Every 2.0s: free -mh                                                                                                                                                                                       relunsec: Fri Jul 17 04:46:12 2026

               total        used        free      shared  buff/cache   available
Mem:           7.0Gi       6.1Gi       182Mi       359Mi       1.0Gi       893Mi
Swap:          4.0Gi       2.3Gi       1.7Gi
```

when running

```css
relunsec@relunsec:~/software/findutils/target/debug$ /bin/find -files0-from /dev/urandom
```

it is not exhausting very fast memory, and outputs the current valid filenames, when pressing CTRL+C. to stop it, the uu find output effectively nothing just consuming and exhausting very fast memory

MsfPablo:
Fixes #779.

## Problem

`parse_files0_args` read the entire `-files0-from` source with `read_to_end` before splitting it on NUL, so `find -files0-from /dev/urandom` (or `/dev/zero`, or any very large list) grew the buffer without bound and never walked a single starting point.

## Fix

The starting points are now produced lazily by a `Files0Paths` iterator built on `BufRead::read_until(0, ..)`, and `do_find` walks each one as soon as it is read. The source is still opened during argument parsing, so `cannot open ... for reading` is still reported before any traversal starts.

Behaviour is otherwise unchanged: a trailing NUL still terminates the last name rather than producing an empty one, zero-length names still warn `invalid zero-length file name` and are skipped, and invalid UTF-8 is still an error — the difference is that it now surfaces when that entry is reached, after the earlier valid entries have been walked, which is closer to the GNU behaviour described in the issue.

Measured on this branch (macOS, `/usr/bin/time -l`):

```
$ /usr/bin/time -l ./target/debug/find -files0-from /dev/urandom
             1884520  peak memory footprint

$ /usr/bin/time -l timeout 3 ./target/debug/find -files0-from /dev/zero > /dev/null
             1048864  peak memory footprint
```

I also checked `src/xargs`: it already reads its input incrementally and shares no code with this path.

## Test

Bounded memory is awkward to assert in the test suite, so the new test covers the observable consequence of streaming: `files0_streams_before_invalid_utf8` feeds `./test_data/simple\0\xff\0` and asserts the valid starting point is printed before the UTF-8 error. That fails on the previous slurp-everything implementation.

## Verification

- `cargo test` — 226 + 60 + 27 + 25 + 15 + 6 passed, 0 failed
- `cargo fmt --check` — clean
- `cargo clippy --all-targets -- -D warnings` — clean

---

Disclosure: this change was prepared with AI assistance (Claude). I reviewed the diff and ran the checks above myself.

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -65,7 +65,6 @@ use ls::Ls;
 use std::{
     error::Error,
     fs::{File, Metadata},
-    io::Read,
     path::Path,
     str::FromStr,
     time::SystemTime,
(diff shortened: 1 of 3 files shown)


MsfPablo:
Quick check on the new bfs failures flagged by the bot:

```
gnu/okdir_path_empty
gnu/okdir_path_relative
```

Both of these are about -`okdir` rejecting -`{exec,ok}dir` when `$PATH` contains a relative entry (matches GNU find behaviour, added in bfs commit `163baf1c9a`). The streaming change here only touches `src/find/mod.rs` and `src/find/matchers/mod.rs` -`files0-from` code paths; `-{ok,exec}dir` lives in `src/find/matchers/exec.rs` and is untouched by this diff.

I checked the latest main-branch artifact (commit aeb8f59730, 314 tests): `gnu/okdir_path_relative` is already listed as FAIL on main, and `gnu/okdir_path_empty` is missing from the older 313-test reference used in this comparison. Running `util/compare_test_results.py` against the freshest reference reduces the regression list to just `gnu/okdir_path_empty` — which is also a pre-existing uutils gap rather than something this PR introduced (the bfs testsuite occasionally skips it depending on the tty / expect environment, which explains why older runs don't list it).

So no fix needed in this PR — the streaming change is doing what it should. The PATH-rejection logic for -`{exec,ok}dir` would be a separate, larger issue (GNU compatibility for `execdir_path_*` is a known gap across the whole uutils findutils suite). Happy to file that as a follow-up if maintainers want.

sylvestre:
these long comments/messages aren't particularly useful, please make them shorter

sylvestre:
> Happy to file that as a follow-up if maintainers want.

sure, thanks

PR #833, opened 2026-08-12
find: avoid panic on a non-UTF-8 argument

### LINKED ISSUE #816: `-printf` panics on an invalid-UTF-8 format string (exit 101)

AlejandroCoronadoN:
## Summary

A non-UTF-8 argument makes `find` panic during argument collection, before the format
parser is ever reached:

```
$ find d -printf $'%\xff|\n'
thread 'main' panicked at ...: called `Result::unwrap()` on an `Err` value
$ echo $?
101
```

## Root cause

`main` collected arguments with `std::env::args()`, which panics on a non-UTF-8 argument.

## Fix

Collect with `std::env::args_os()` and report the first non-UTF-8 argument as an error,
exiting with code 1 instead of aborting. The conversion is extracted into a small
`collect_utf8_args` helper so it can be unit-tested. Full GNU-style byte pass-through of
the format string is a larger change and left as a follow-up.

## Verification

```
$ find d -printf $'%\xff|\n'
find: invalid (non-UTF-8) argument: %<0xff>|
$ echo $?
1
```

Adds unit tests `collects_valid_utf8_args` and `non_utf8_arg_is_rejected_not_panicking`.
`cargo fmt` and `cargo clippy --bin find` are clean.

Fixes #816.

### DIFF
--- a/src/find/main.rs
+++ b/src/find/main.rs
@@ -4,14 +4,59 @@
 // license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
+use std::ffi::OsString;
+
+/// Collect the process arguments as UTF-8 strings.
+///
+/// Unlike `std::env::args`, which panics on a non-UTF-8 argument, this reads the
+/// arguments with `args_os` and reports the first invalid one as an error so the
+/// caller can exit gracefully instead of aborting (#816).
+fn collect_utf8_args(args: impl Iterator<Item = OsString>) -> Result<Vec<String>, OsString> {
+    args.map(OsString::into_string).collect()
+}
+
 fn main() {
     // Ignores the SIGPIPE signal.
     // This is to solve the problem that when find is used with a pipe character,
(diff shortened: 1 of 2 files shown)


sylvestre:
please add a test_find.rs

AlejandroCoronadoN:
Done, added an integration test in test_find.rs that passes a non-UTF-8 argument and checks it errors cleanly instead of panicking.

PR #835, opened 2026-08-13
find: reject a signed -maxdepth/-mindepth value to match GNU

AlejandroCoronadoN:
## Problem

`-maxdepth`/`-mindepth` accepted a `+`-signed value that GNU rejects:

```
$ find dir -maxdepth +1     # uutils: runs normally (parsed as 1)
$ gfind dir -maxdepth +1
find: Expected a positive decimal integer argument to -maxdepth, but got '+1'
```

Rust's `usize::from_str` accepts a leading `+`, so `convert_arg_to_number` let
it through, which is inconsistent with the function's own "positive decimal
integer" error message.

## Fix

Only accept plain decimal digits.

## Verification

Compared against GNU findutils (`gfind`) for `-maxdepth` and `-mindepth` over
`1`, `+1`, `-1`, `+2`, `0`, `2`; exit codes match in every case. Added a unit
test; `cargo test --lib` passes and `cargo fmt` / `cargo clippy` are clean.

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -307,8 +307,10 @@ fn convert_arg_to_number(
     option_name: &str,
     value_as_string: &str,
 ) -> Result<usize, Box<dyn Error>> {
+    // Only accept plain decimal digits. Rust's `usize::from_str` also accepts a
+    // leading '+', but GNU find rejects a signed value like "+1" here.
     match value_as_string.parse::<usize>() {
-        Ok(val) => Ok(val),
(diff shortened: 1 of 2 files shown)


sylvestre:
please add a test_find.rs

AlejandroCoronadoN:
Done, added an integration test in test_find.rs for the signed -maxdepth/-mindepth case.

sylvestre:
a bunch of jobs have been failing

AlejandroCoronadoN:
The failing jobs were from a stale base; rebased the branch on the latest main. fmt, clippy and the tests pass locally now.

PR #839, opened 2026-08-17
find: add -warn and -nowarn options

kevinburke:
GNU find distinguishes warnings about inadvisable command-line usage from errors encountered while traversing directories. These warnings do not change find's exit status. By default, GNU enables them only when standard input is a terminal and POSIXLY_CORRECT is unset; otherwise it disables them so existing scripts do not gain unsolicited diagnostics.

The -warn and -nowarn options change that state at the point where each appears. Thus `-warn -type d -maxdepth 1` warns when parsing reaches -maxdepth, while placing -nowarn before -maxdepth suppresses the warning. Putting -nowarn at the end cannot retract a warning already emitted. GNU leaves the active warnings unspecified when POSIXLY_CORRECT and an explicit -warn are both present.

Global options such as -maxdepth are semantically non-positional: they affect tests before and after their location. GNU accepts a global option after a test or action, but warns that the ordering is misleading when warnings are enabled.

uutils previously rejected -warn and -nowarn as unknown predicates. Track the current warning state and most recent test or action, accept both options, and implement GNU's misplaced-global-option diagnostic for the supported global options. GNU also controls warnings for deprecated -d and slashes in -name or -iname patterns; those remain outside this change.

Add coverage for enabling the diagnostic and disabling it before the global option is parsed.

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -447,6 +447,12 @@ fn get_or_create_file(path: &str) -> Result<File, Box<dyn Error>> {
     Ok(file)
 }
 
+/// A "global option" (GNU find's term) applies to the whole command instead
+/// of a specific test, so unlike a test it takes effect no matter where in
+/// the expression it appears; `-maxdepth` after `-type f`, for instance,
+/// still limits the depth of every path find visits, not just descendants of
+/// paths matching `-type f`. GNU find warns when one appears after a test, as
+/// implemented by `warn_if_global_option_follows_test` below.
 fn is_global_option(argument: &str) -> bool {
     matches!(
         argument,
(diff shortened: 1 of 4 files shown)


kevinburke:
The latest change should have fixed the test suite issues. Github is having some problems, so not clear the automation will run correctly.

sylvestre:

src/find/matchers/mod.rs
```diff
@@ -447,6 +447,62 @@ fn get_or_create_file(path: &str) -> Result<File, Box<dyn Error>> {
     Ok(file)
 }
 
+fn is_global_option(argument: &str) -> bool {
```
sylvestre, line 456: what is a global option?

kevinburke: Added a doc explaining what this means (it's a GNU term)

PR #841, opened 2026-08-17
find: match GNU rounding for -amin/-cmin/-mmin

### LINKED ISSUE #813: find: find -amin 1 output differs with GNU
uutils findutils:
```
$ cargo run --bin find -- -amin 1
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/find -amin 1`
```
GNU findutils:
```
$ find -amin 1
./target/debug/find.d
```

MsfPablo:
Fixes #813.

`-amin`/`-cmin`/`-mmin` computed the age in minutes with a truncating division, so a file accessed a few seconds ago landed in bucket 0 and `find -amin 1` didn't list it, while GNU does.

I checked the actual behaviour against GNU findutils 4.11 rather than guessing, and it turns out the three comparator forms don't share one rounding rule:

- `-amin n` matches when `(n-1)*60 <= age < n*60`, i.e. an age of `d` seconds falls in bucket `d/60 + 1`
- `-amin -n` matches when `age < n*60`
- `-amin +n` matches when `age >= n*60`

Feeding a single `age_in_minutes` into `ComparableValue` can't express all three at once (whichever value you pick, one of the forms is off by one), so the comparison is now done against the age in seconds directly.

Verified against GNU on files aged 0/30/59/60/61/90/119/120/121/179/180/181 seconds — output is now identical for `-mmin` with 0, 1, 2, 3, -1, -2, +0, +1 and +2. Added a unit test pinning those boundaries; the two existing `file_age_range_matcher` assertions still pass unchanged.

### DIFF
--- a/src/find/matchers/time.rs
+++ b/src/find/matchers/time.rs
@@ -383,8 +383,23 @@ impl FileAgeRangeMatcher {
             }
         };
         let age_in_seconds: i64 = age.as_secs() as i64 * if is_negative { -1 } else { 1 };
-        let age_in_minutes = age_in_seconds / 60 + if is_negative { -1 } else { 0 };
-        Ok(self.minutes.imatches(age_in_minutes))
+
+        // GNU find treats `-amin n` as "the age falls in the n-th minute", i.e.
+        // an age of d seconds matches exactly n when (n-1)*60 <= d < n*60. So a
+        // file accessed a few seconds ago matches `-amin 1`, not `-amin 0`.
+        // `-amin -n` and `-amin +n` compare the age against n*60 seconds
+        // directly, which is not the same off-by-one as the exact case.
+        Ok(match self.minutes {
+            ComparableValue::EqualTo(n) => {
+                age_in_seconds >= 0 && (age_in_seconds as u64) / 60 + 1 == n
+            }
+            ComparableValue::LessThan(n) => {
+                age_in_seconds < 0 || (age_in_seconds as u64) < n.saturating_mul(60)
+            }
+            ComparableValue::MoreThan(n) => {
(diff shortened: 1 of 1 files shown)


MsfPablo:
The bfs job flags `common/amin` and `common/mmin` as new failures. That is this change, and I think it is the right side of the difference — bfs and GNU disagree here, and `tests/common/amin.out` encodes bfs's answer.

Same tree bfs's test builds (files at -3600s, -121s, -61s, -30s, +60s, +3600s), run against GNU findutils 4.11, bfs 4.1.4, and this branch:

```
-1  gfind     in_one_hour in_one_minute thirty_seconds_ago
-1  bfs       in_one_hour in_one_minute thirty_seconds_ago
-1  ours      in_one_hour in_one_minute thirty_seconds_ago

1   gfind     thirty_seconds_ago
1   bfs       one_minute_ago
1   ours      thirty_seconds_ago

+1  gfind     one_hour_ago one_minute_ago two_minutes_ago
+1  bfs       one_hour_ago two_minutes_ago
+1  ours      one_hour_ago one_minute_ago two_minutes_ago
```

The bfs rows are exactly what `common/amin.out` contains, so those two tests are asserting bfs behaviour rather than GNU's. bfs buckets an age of 61s as minute 1; GNU buckets it as minute 2 and puts the 30s file in minute 1. `Run GNU findutils tests` passes, and the same job reports `gnu/files0_from_ok` and `gnu/okdir_path_empty` moving to pass, so the net FAILED count is unchanged.

Happy to add these two to `ignore-intermittent.txt` (or wherever known bfs divergences belong) if that is how you would rather record it — I did not want to touch that list unprompted.

(Reposting: my earlier comment had stray backslashes from a shell quoting slip.)

PR #842, opened 2026-08-17
printf: format %S like GNU (%g, not one decimal)

MsfPablo:
I was diffing our `find` against GNU findutils 4.11 across the `-printf` directives and `%S` came out differently on every file that isn't a round ratio.

GNU prints the sparseness ratio with `%g`; we use `{:.1}`:

```
$ LC_ALL=C gfind . -printf '%S %s %b %p\n' | sort      $ LC_ALL=C find . -printf '%S %s %b %p\n' | sort
0 224 0 .                                               0.0 224 0 .
1 0 0 ./empty                                           1.0 0 0 ./empty
682.667 6 8 ./sub/b.log                                 682.7 6 8 ./sub/b.log
1365.33 3 8 ./a.txt                                     1365.3 3 8 ./a.txt
2048 2 8 ./sub/deep/c.md                                2048.0 2 8 ./sub/deep/c.md
```

So `%g` differs from `{:.1}` in three ways that all show up here: 6 significant digits rather than 1 decimal place, trailing zeros and a trailing `.` stripped, and a switch to exponent form outside `[1e-4, 1e6)`.

Rust has no `%g`, so this adds a small `format_g` helper and routes `%S` through it. The zero-length and non-unix branches returned the literal `"1.0"` and are now `"1"`, matching the `./empty` line above.

The unit test's expected values are the output of C's `printf("%g", ...)` under glibc with `LC_ALL=C`, including the rounding-changes-the-exponent case (`999999.9` -> `1e+06`) and the exponent-form case (`1234567` -> `1.23457e+06`). After the change `find -printf '%S %p\n'` is byte-identical to GNU on my test tree.

One thing I deliberately left alone: GNU honours `LC_NUMERIC` for `%S` (it prints `1365,33` in a comma-decimal locale) while we always emit `.`. That's a separate matter from the formatting itself, so it felt out of scope here.

### DIFF
--- a/src/find/matchers/printf.rs
+++ b/src/find/matchers/printf.rs
@@ -20,6 +20,45 @@ use std::os::unix::prelude::MetadataExt;
 #[cfg(unix)]
 const STANDARD_BLOCK_SIZE: u64 = 512;
 
+/// Formats `value` the way C's `printf("%g", value)` would.
+///
+/// GNU find prints `%S` with `%g`, so a ratio of exactly 2048 comes out as
+/// `2048` rather than `2048.0`, and 512*8/6 as `682.667` rather than `682.7`.
+#[cfg(unix)]
+fn format_g(value: f64) -> String {
+    /// The default precision of `%g`, in significant digits.
+    const PRECISION: i32 = 6;
+
+    fn trim(s: &str) -> &str {
+        match s.split_once('.') {
+            Some(_) => s.trim_end_matches('0').trim_end_matches('.'),
+            None => s,
+        }
+    }
+
+    if !value.is_finite() || value == 0.0 {
+        return format!("{value}");
+    }
+
+    // Round to PRECISION significant digits first, so that a value such as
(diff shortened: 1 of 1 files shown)


src/find/matchers/printf.rs
```diff
@@ -20,6 +20,45 @@ use std::os::unix::prelude::MetadataExt;
 #[cfg(unix)]
 const STANDARD_BLOCK_SIZE: u64 = 512;
 
+/// Formats `value` the way C's `printf("%g", value)` would.
+///
+/// GNU find prints `%S` with `%g`, so a ratio of exactly 2048 comes out as
+/// `2048` rather than `2048.0`, and 512*8/6 as `682.667` rather than `682.7`.
+#[cfg(unix)]
```
sylvestre, line 27: why only unix ?

sylvestre: @MsfPablo ping?

sylvestre:
@MsfPablo last ping before i close it

PR #845, opened 2026-08-19
find: warn on -name/-path patterns with a directory separator

### LINKED ISSUE #783: bug(find): do not warn when appending / to -name in the start or the end
Hi, findutils mainteners

we confirmed uu find do not emit zero single warning

```css
relunsec@relunsec:~/software/findutils/target/debug$ ./find -name /a.txt
relunsec@relunsec:~/software/findutils/target/debug$ /bin/find -name /a.txt
/bin/find: warning: ‘-name’ matches against basenames only, but the given pattern contains a directory separator (‘/’), thus the expression will evaluate to false all the time.  Did you mean ‘-wholename’?
```

while the gnu find those

the samething with

```css
relunsec@relunsec:~/software/findutils/target/debug$ ./find -wholename a.txt/
relunsec@relunsec:~/software/findutils/target/debug$ /bin/find -wholename a.txt/
/bin/find: warning: -wholename a.txt/ will not match anything because it ends with /.
```

the samething with 

```css
relunsec@relunsec:~/software/findutils/target/debug$ /bin/find -path ./
/bin/find: warning: -path ./ will not match anything because it ends with /.
```

not just name

```css
relunsec@relunsec:~/software/findutils/target/debug$ ./find -path ./
```

MsfPablo:
Closes #783.

GNU `find` emits a warning when:

- a `-name`/`-iname` pattern contains a `/` — `-name` matches basenames only, so such a pattern can never match and the user likely meant `-wholename`;
- a `-path`/`-ipath`/`-wholename`/`-iwholename` pattern ends with `/` — a trailing separator can never match a real path.

This PR makes `find` emit the same warnings on stderr while still exiting successfully.

```console
$ find -name a/b
find: warning: '-name' matches against basenames only, but the given pattern contains a directory separator ('/'), thus the expression will evaluate to false all the time.  Did you mean '-wholename'?
$ find -path a/
find: warning: -path a/ will not match anything because it ends with /.
```

**All-slash exemption:** a pattern made up only of `/` (e.g. `-name /` or `-path /`) is left alone — `/` is a valid basename for the root entry, so warning would be a false alarm. GNU itself miscategorises this case (GNU bug [#62227](https://lists.nongnu.org/archive/html/bug-findutils/2022-03/msg00004.html)); the existing `find_slashes` test encodes the correct behaviour and continues to pass.

**Tests:** four new regression tests in `tests/test_find.rs` cover the warning cases (`name_pattern_with_separator_warns`, `path_pattern_ending_with_separator_warns`), the no-warning positive case (`name_pattern_without_separator_does_not_warn`), and the all-slash exemption (`all_slash_pattern_does_not_warn`). `cargo clippy -p findutils --tests -- -D warnings` and `cargo fmt --check` are clean; the full `test_find` suite passes (73 passed, 0 failed).

### DIFF
--- a/src/find/matchers/mod.rs
+++ b/src/find/matchers/mod.rs
@@ -65,6 +65,7 @@ use ls::Ls;
 use std::{
     error::Error,
     fs::{File, Metadata},
+    io::{stderr, Write},
     path::{Path, PathBuf},
     str::FromStr,
     time::SystemTime,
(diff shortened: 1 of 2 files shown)


src/find/matchers/mod.rs
```diff
@@ -534,13 +535,40 @@ fn build_matcher_tree(
                     return Err(From::from(format!("missing argument to {}", args[i])));
                 }
                 i += 1;
+                // GNU find warns when a -name/-iname pattern contains a directory
```
sylvestre, line 538: does it really need a 6 lines comment ? please make it shorter

PR #848, opened 2026-08-21
find: accept egrep and posix-egrep regex types to match GNU

AlejandroCoronadoN:
GNU `find` accepts `egrep` and `posix-egrep` as `-regextype` names; both select the extended (ERE) syntax. uutils rejects them:

```
$ find . -regextype egrep -regex '.*'
find: Invalid regex type: egrep (must be one of 'emacs', 'grep', 'posix-basic', 'posix-extended')
```

They behave identically to the already-supported `posix-extended`. Verified against GNU with a `+` quantifier: `egrep`, `posix-egrep` and `posix-extended` all match the same paths. This maps both aliases to `PosixExtended`, mirroring how `ed`/`sed` already alias `posix-basic`.

Left out on purpose: GNU's `awk`, `gnu-awk`, `posix-awk` and `posix-minimal-basic` have small semantic differences that the onig syntaxes do not represent exactly, so they are not added here.

### DIFF
--- a/src/find/matchers/regex.rs
+++ b/src/find/matchers/regex.rs
@@ -70,6 +70,8 @@ impl FromStr for RegexType {
             "posix-extended" => Ok(Self::PosixExtended),
             // ed and sed are the same as posix-basic
             "ed" | "sed" => Ok(Self::PosixBasic),
+            // egrep and posix-egrep are the same as posix-extended
+            "egrep" | "posix-egrep" => Ok(Self::PosixExtended),
             _ => Err(ParseRegexTypeError(s.to_owned())),
         }
     }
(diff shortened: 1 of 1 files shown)


AlejandroCoronadoN:
Closing this. The bfs test `gnu/regextype_egrep` shows that GNU's `egrep` syntax is not equivalent to `posix-extended`: GNU accepts patterns like `*.*/{l` (a leading `*` and an incomplete `{` are treated as literals), while `posix-extended` (Oniguruma) rejects them with "target of repeat operator is not specified". So mapping `egrep`/`posix-egrep` onto `PosixExtended` is not correct parity. A proper fix needs an egrep-equivalent syntax, which is more involved. Sorry for the noise.

PR #849, opened 2026-08-21
find: do not double the "find:" prefix on an invalid -type list

AlejandroCoronadoN:
A `-type`/`-xtype` list that ends in a comma prints the `find:` prefix twice:

```
$ find . -type f,
find: find: Last file type in list argument to -type is missing, i.e., list is ending on: ','
```

GNU prints it once:

```
$ find . -type f,
find: Last file type in list argument to -type is missing, i.e., list is ending on: ','
```

The message in `type_creator` hardcodes a leading `find: `, but the top level error printer already prepends `find: {e}`. The two sibling messages in the same function (the duplicate type and the missing separator ones) do not hardcode it, so this one is inconsistent. Drop the hardcoded prefix so the output matches GNU and the neighbors.


cakebaker:
Thanks!

PR #851, opened 2026-08-21
find: reject the deprecated -perm +MODE octal form to match GNU

AlejandroCoronadoN:
GNU `find` rejects the deprecated `-perm +MODE` octal form (a leading `+` followed by an octal number). uutils silently accepts it: the `+` is treated as an unknown prefix, so it falls through to an exact numeric comparison and matches files instead of erroring.

```
$ find . -perm +644 ; echo $?
./f          # wrongly matches
0
$ gfind . -perm +644 ; echo $?
gfind: invalid file mode '+644'
1
```

Same for `+0` and `+400`. Symbolic forms (`+rwx`, `+u+w`) and the `644` / `/644` / `-644` forms are unaffected. This rejects a leading `+` followed by a digit, matching GNU.

### DIFF
--- a/src/find/matchers/perm.rs
+++ b/src/find/matchers/perm.rs
@@ -53,6 +53,13 @@ mod parsing {
     }
 
     pub fn parse_mode(pattern: &str, for_dir: bool) -> Result<u32, Box<dyn Error>> {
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #857, opened 2026-09-01
fix(find): output nothing when -mindepth exceeds -maxdepth

### LINKED ISSUE #778: bug(find): when -mindepth greater than -maxdepth it outputs files, even if must not

Jorge-Polanco-Roque:
Fixes #778

## Problem

When `-mindepth` is greater than `-maxdepth`, `find` should output nothing (no depth can be both `>= mindepth` and `<= maxdepth`), matching GNU find. uutils instead printed the max-depth entries:

```console
$ find . -mindepth 2 -maxdepth 1     # GNU: (nothing)
$ ./find . -mindepth 2 -maxdepth 1   # uutils, before:
./a
./f0
```

## Cause

`process_dir` passes both bounds to `WalkDir::min_depth`/`max_depth`. When `min_depth > max_depth`, `WalkDir` does not yield the empty set — it effectively clamps the range and still yields entries at the max depth, so they leak through.

## Fix

Short-circuit in `process_dir`: when `config.min_depth > config.max_depth`, return early with no output, matching GNU find.

## Tests

Adds `find_mindepth_greater_than_maxdepth`, which asserts empty output for `-mindepth 2 -maxdepth 1`. It **fails without the fix** (the max-depth entries are printed) and passes with it. `cargo fmt --check` and `cargo clippy` are clean; normal min/max combinations are unaffected.

### DIFF
--- a/src/find/mod.rs
+++ b/src/find/mod.rs
@@ -276,6 +276,13 @@ fn process_dir(
     matcher: &dyn matchers::Matcher,
     quit: &mut bool,
 ) -> i32 {
+    // No depth can be both >= min_depth and <= max_depth when min_depth exceeds
+    // max_depth, so nothing matches — the same as GNU find. `WalkDir` would
+    // otherwise still yield the max-depth entries, so short-circuit here.
+    if config.min_depth > config.max_depth {
+        return 0;
+    }
+
     let mut walkdir = WalkDir::new(dir)
         .contents_first(config.depth_first)
         .max_depth(config.max_depth)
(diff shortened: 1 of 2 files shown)


sylvestre:
please add a test in tests/test_find.rs too :)

Jorge-Polanco-Roque:
Done — added an integration test in `tests/test_find.rs` (`mindepth_exceeds_maxdepth_outputs_nothing`) that runs `-mindepth 2 -maxdepth 1` against `test_data/simple` and asserts empty stdout + success, matching GNU find. Verified it fails without the guard and passes with it. :)

Jorge-Polanco-Roque:
Added the integration test in tests/test_find.rs (`mindepth_exceeds_maxdepth_outputs_nothing`), it's up on the branch. Happy to cover more cases if you'd like.

cakebaker:
Thanks for your PR!

PR #861, opened 2026-09-10
find, locate: replace onig with fancy-regex

### LINKED ISSUE #212: consider making onig optional
There are a number of pure rust globs and regex crates, might be better to not have to deal with a C dependency if we could avoid it.

From what I'm seeing `regex` may lack a mean to select a specific flavour of regex, not sure if somebody already had a mean to restrict the engine to not support extensions compared to posix/emacs.

wtcpython:
Closes #212

### DIFF
--- a/src/find/matchers/regex_transpile.rs
+++ b/src/find/matchers/regex_transpile.rs
@@ -1,0 +1,187 @@
+// This file is part of the uutils findutils package.
+//
+// For the full copyright and license information, please view the LICENSE
+// file that was distributed with this source code.
+
+use std::error::Error;
+
+use fancy_regex::{Regex, RegexBuilder};
+
+use super::regex::RegexType;
+
+/// Swap BRE escapes to fancy-regex syntax (`\(` → `(`, `+` → `\+`, etc.).
+fn transpile_bre_like(pattern: &str) -> String {
+    let mut output = String::with_capacity(pattern.len() * 2);
+    let mut chars = pattern.chars().peekable();
+    let mut in_bracket = false;
+
+    while let Some(ch) = chars.next() {
+        if in_bracket {
+            if ch == ']' && output.ends_with(|c| c != '\\' && c != '[' && c != '^') {
+                in_bracket = false;
+            }
+            output.push(ch);
+            continue;
+        } else if ch == '[' {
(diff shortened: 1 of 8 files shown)


src/find/matchers/regex_transpile.rs
```diff
@@ -0,0 +1,180 @@
...
+                        output.push_str(r"\$");
+                    }
+                }
+                _ => output.push(ch),
+            }
+            continue;
+        }
+
+        match chars.next() {
+            Some('(') => output.push('('),
```
oech3: Something like `c if "(){}|+?".contains(c) => output.push(c)`

src/find/matchers/regex_transpile.rs
```diff
@@ -0,0 +1,185 @@
...
+///
+/// Emacs uses `\(` `\)` `\{m,n\}` (BRE-style escapes) for groups and
+/// intervals, but ERE-style `+` `?` `|` for quantifiers and alternation.
+fn transpile_emacs(pattern: &str) -> String {
+    let mut output = String::with_capacity(pattern.len() + 8);
+    let mut chars = pattern.chars().peekable();
+
+    while let Some(ch) = chars.next() {
+        if ch != '\\' {
+            if matches!(ch, '(' | ')' | '{' | '}') {
```
oech3: `b"(){}".contains(ch)`

oech3:
drop https://github.com/uutils/findutils/blob/fda8855b739fae55cf80e94e2d46f991d8a58c2a/.github/workflows/ci.yml#L74-L75

lhecker:
I've been made aware of the series of changes for uutils to migrate away from onig via https://github.com/uutils/grep/pull/117. I'd like to also mention here, that this PR regresses matches that rely on leftmost longest semantics.

Example:
```sh
touch a ab
find a ab -regex 'a\|z'
```

keith-hall:
note that I have a PR prepared for fancy-regex to enable leftmost longest match mode: https://github.com/fancy-regex/fancy-regex/pull/281

sylvestre:
cool, thanks :)

oech3:
@wtcpython reopen?

wtcpython:
Reopened: #864

PR #864, opened 2026-09-18
find, locate: replace onig with fancy-regex

### LINKED ISSUE #212: consider making onig optional

wtcpython:
Closes #212


oech3:
> Merging this PR will degrade performance by 36.61%

Oh no...

Cargo.toml
```diff
@@ -22,9 +22,9 @@ chrono = "0.4.45"
 clap = { version = "4.6", features = ["env"] }
 faccess = "0.2.4"
 nix = { version = "0.31", features = ["fs", "user"] }
-onig = { version = "6.5", default-features = false }
+fancy-regex = { version = "0.19.2", default-features = false, features = ["std", "unicode"] }
 regex = "1.12"
-uucore = { version = "0.10.0", features = ["entries", "fs", "fsext", "mode"] }
+uucore = { git = "https://github.com/wtcpython/coreutils.git", branch = "uucore-regex", features = ["entries", "fs", "fsext", "mode", "regex"] }
```
oech3, line 27: Oops. uucore 0.12.0 was just released about 2 days ago. New version would not appear soon...

oech3: We can copy & paste code and add todo comment with linking your PR to coreutils, and drop copied code when new uucore was released.

sylvestre:
the ci is red :)

src/find/matchers/regex.rs
```diff
@@ -85,30 +85,27 @@ impl RegexMatcher {
         pattern: &str,
         ignore_case: bool,
     ) -> Result<Self, Box<dyn Error>> {
-        let syntax = match regex_type {
-            RegexType::Emacs => Syntax::emacs(),
-            RegexType::Grep => Syntax::grep(),
-            RegexType::PosixBasic => Syntax::posix_basic(),
-            RegexType::PosixExtended => Syntax::posix_extended(),
+        // GNU find's -regex does full-path matching, so anchor the pattern.
+        let anchored = if pattern.starts_with('^') && pattern.ends_with('$') {
```
sylvestre, line 89: anchoring by concatenation breaks alternation: `a|b` becomes `^a|b$`, which means `^a` or `b$`. please wrap it instead, something like `^(?:{pattern})$`

src/find/matchers/regex.rs
```diff
@@ -85,30 +85,27 @@ impl RegexMatcher {
...
+            pattern.to_owned()
+        } else if pattern.starts_with('^') {
+            format!("{pattern}$")
+        } else if pattern.ends_with('$') {
+            format!("^{pattern}")
+        } else {
+            format!("^{pattern}$")
```
sylvestre, line 96: this changes -regex/-name matching semantics, please add tests in tests/test_find.rs (alternation, ^/$ in the middle, -iregex, each -regextype)

src/find/matchers/regex.rs
```diff
@@ -85,30 +85,27 @@ impl RegexMatcher {
...
+            regex: regex_transpile::compile(&anchored, regex_type, ignore_case)?,
+        })
     }
 }
 
 impl Matcher for RegexMatcher {
     fn matches(&self, file_info: &WalkEntry, _: &mut MatcherIO) -> bool {
         self.regex
             .is_match(file_info.path().to_string_lossy().as_ref())
+            .unwrap_or(false)
```
sylvestre, line 108: unwrap_or(false) silently turns a backtrack-limit error into "no match", is that what we want?

src/find/matchers/regex_transpile.rs
```diff
@@ -0,0 +1,104 @@
...
+        if ch != '\\' {
+            if "(){}".contains(ch) {
+                output.push('\\');
+            }
+            output.push(ch);
+            continue;
+        }
+
+        match chars.next() {
+            Some(c) if "(){}|+?".contains(c) => output.push(c),
```
sylvestre, line 30: in emacs syntax a bare `|` is a literal and only `\|` is alternation, so `|` needs to be escaped in the other branch. the doc comment above says the opposite

src/find/matchers/glob.rs
```diff
@@ -158,20 +154,22 @@ pub struct Pattern {
...
-        };
-
         // As long as glob_to_regex() is correct, this should never fail
-        let regex = glob_to_regex(pattern).map(|r| parse_bre(&r, options).unwrap());
+        let regex = glob_to_regex(pattern).map(|r| {
+            // GNU find's -path, -name, -lname etc. do full-string matching,
+            // so anchor the regex to prevent partial matches.
+            let anchored = format!("^{r}$");
+            regex_transpile::compile(&anchored, super::regex::RegexType::PosixBasic, caseless)
+                .unwrap()
```
sylvestre: unwrap() on a user-supplied pattern, could you please propagate the error? the "should never fail" comment was true for parse_bre, not for the new transpiler

Issue #164, opened 2022-05-25
xargs doesn't limit command line lengths properly


tavianator:
```console
$ find ~ -print0 | ./target/debug/xargs -0 echo >/dev/null
Error: Command could not be run: Argument list too long (os error 7)
```

From a quick look, there's at least an issue here: https://github.com/uutils/findutils/blob/36e3229537de84a9b68bedc984a4366a6ea24357/src/xargs/mod.rs#L141-L146

This needs to count not only the length of the string, but also the size of the pointer that ends up in `argv`/`envp`.

It might be worth using the https://crates.io/crates/argmax crate to do this accounting for us.  It could also help for #6.

Every2:
Hi, @tavianator . What you think about use something like:
```rust
#[cfg(unix)]
fn count_osstr_chars_for_exec(s: &OsStr) -> (usize, usize) {
    use std::os::unix::ffi::OsStrExt;
    let cmd = Command::new(s);
    let args: Vec<&OsStr> = cmd.get_args().collect();
    // Include +1 for the null terminator.
    (s.as_bytes().len() + 1, args.len())
}
```
To this issue? We mantain the count for string and get argv count together.

tavianator:
@Every2 I would strongly recommend using https://crates.io/crates/argmax, it already contains a well-tested implementation of command line length limiting.  It was also recently added to this project's `find` implementation: #515

Every2:
@tavianator I see. Just confirming if I understood correctly. The structure of the code shouldn't be changed, only the function `count_osstr_chars_for_exec` instead of using an OsStr, use argmax equivalent (I believe argmax::command and it's methods)?

tavianator:
> The structure of the code shouldn't be changed, only the function `count_osstr_chars_for_exec`

No, i feel like the code in xargs/mod.rs should be mostly re-written to take advantage of the argmax crate instead of using its own logic (MaxArgsCommandSizeLimiter etc.)

Every2:
> No, i feel like the code in xargs/mod.rs should be mostly re-written to take advantage of the argmax crate instead of using its own logic (MaxArgsCommandSizeLimiter etc.)

I'll start working on it then! One more thing. Is it better to send small PRs or send it all at once?

PR #867, opened 2026-09-20
xargs: account for argv and envp pointer overhead

ddy314:
The system command-size limiter counts string bytes but misses argv/envp pointers. With a large explicit `-s`, many short arguments fit the character budget yet fail to execute with E2BIG.

Account for argument and environment pointers in the system limiter. The user-specified `-s` limit still counts string bytes, including terminators. This addresses the pointer-accounting case in #164.

Validation: the new Linux regression fails with E2BIG before the fix and passes afterward, preserving every argument across multiple invocations. All 34 xargs integration tests and 15 xargs unit tests pass, along with `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings`.

AI-assisted implementation and local testing.


src/xargs/mod.rs
```diff
@@ -226,15 +228,19 @@ impl MaxCharsCommandSizeLimiter {
             .map(|(var, value)| count_osstr_chars_for_exec(var) + count_osstr_chars_for_exec(value))
             .sum();
 
-        Self::new(arg_max.saturating_sub(ARG_HEADROOM + env_size))
+        let pointer_size = std::mem::size_of::<*const uucore::libc::c_char>();
```
sylvestre: `size_of::<*const c_char>()` shows up 3 times (here and in both tests), a small const would do, no?

src/xargs/mod.rs
```diff
@@ -1366,6 +1372,28 @@ mod tests {
         assert!(limiter.try_arg(make_arg_hard("a"), empty_cursor()).is_ok());
     }
 
+    #[cfg(unix)]
+    #[test]
+    fn test_system_limiter_counts_empty_argument_pointers() {
+        let mut limiter = MaxCharsCommandSizeLimiter::new_system(&HashMap::new());
+        let size = 1 + std::mem::size_of::<*const uucore::libc::c_char>();
+        limiter.max_chars = size;
```
sylvestre: poking max_chars directly is a bit hacky, could new_system take arg_max as a parameter instead?

tests/test_xargs.rs
```diff
@@ -262,6 +262,23 @@ fn xargs_explicit_size_can_exceed_default_cap() {
     );
 }
 
+#[test]
+#[cfg(target_os = "linux")]
+fn xargs_explicit_size_accounts_for_argument_pointers() {
+    let arg_max = unsafe { uucore::libc::sysconf(uucore::libc::_SC_ARG_MAX) } as usize;
```
sylvestre, line 268: why only linux? the limiter is cfg(unix)

src/xargs/mod.rs
ddy314: Added a cfg(unix) POINTER_SIZE constant and reused it in the implementation and tests in 6307c77.

src/xargs/mod.rs
ddy314: new_system now takes arg_max as a parameter, so the unit test constructs the exact budget without mutating max_chars in 6307c77.

tests/test_xargs.rs
ddy314: Changed the integration test guard to cfg(unix) in 6307c77. The full test suite and all-feature Clippy pass.

ddy314:
The remaining Ubuntu test and coverage failures are unrelated to this PR: both fail in find::matchers::printf::tests::test_parse_multibyte_char_after_directive, while the new xargs tests pass. The same failure is present on main in run https://github.com/uutils/findutils/actions/runs/35599345316.

